Issue #41, opened 2015-09-13
$.prop()


obuchtala:
What shall we do with this part of jquery? http://api.jquery.com/prop/
I fear the confusion adding this would create.

obuchtala:
$.prop is for html properties such as 'type' of an 'input'. So far we did not need this often, or even not at all. Thus, I would rather deviate from jquery here, and call this `htmlProp()`.

```
render() {
  return $$('input').htmlProp('type', 'text');
}
```

PR #42, opened 2015-09-13
Html props

obuchtala:
Implements #41

### DIFF
--- a/test/unit/ui/component.test.js
+++ b/test/unit/ui/component.test.js
@@ -55,7 +55,7 @@ QUnit.test("Different mount scenarios", function(assert) {
   comp = Component.mount($$(SimpleComponent), $('#qunit-fixture'));
   assert.equal(comp.didMount.callCount, 1, "didMount must not be called when mounting to attached elements");
   assert.equal(comp.didRender.callCount, 1, "didRender must have been called once");
-  assert.ok(comp.$el);  
+  assert.ok(comp.$el);
 
   // Mount, passing a Component instance instead of a VirtualComponent
   comp = new SimpleComponent("root");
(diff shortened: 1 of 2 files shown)


michael:
Nice for comp.text(). I have to admit I've been always confused about props vs attrs in jQuery. I just wonder what are situations where attr does not suffice for setting things?

``` js
QUnit.test("Accessing a HTML property via comp.htmlProp()", function(assert) {
  var comp = $$('input').htmlProp('type', 'text')._render();
  assert.equal(comp.htmlProp('type'), 'text', 'Input field should be of type "text".');
});
```

This would work fine with with attr too?

I wonder if we should only implement a read only htmlProp() version for accessing things like `nodeType` which don't have an attribute correspondence.

Generally speaking, for our $$ constructor we either construct a user component by passing props (some app data) or we construct an html element, where we pass html attributes, and css classes etc. just enough to construct an html element in the end. I'm not sure if it needs the difference between attrs and props (for settings things i mean).

I'm just thinking how we can avoid confusion between user component props, html attributes and html props.

If I'm calling attr('data-id', 'foo') on a user component, this attr data gets stored on the element right? not on the component (props).

obuchtala:
The concern about confusion was the centre of my question in #41.

Actually, input.value is one example of a property. $.attr('value', val) sets the default value, but on the other side $.attr('value') does not provide the correct value after being updated by user. Looking at the native DOM element the difference is, that $.attr() changes el.attributes, while $.prop() the property of el.

So there are actually 2 different concepts, while as I said, props are way less often used than attributes, because most often we are using spans and divs which don't have any props we need.

obuchtala:
You're right about input.type, it can be set using $.attr().

obuchtala:
See http://api.jquery.com/prop/, there is an explanation of the difference between attr and prop.

michael:
We should have some dedicated naming discussion I think.. and when we want to be jQuery compatible with our API. It's good to have support for props. Let's merge this and discuss later how we can reduce confusion. The props vs attrs dilemma is definitely not our fault. ;)

PR #44, opened 2015-09-13
Numerical addressing 

obuchtala:
Moving towards a next generation internal addressing scheme which will hopefully help us getting rid of the ugly and error prone container + container annotation index implementation.

### DIFF
--- a/document/container.js
+++ b/document/container.js
@@ -84,7 +84,7 @@ Container.Prototype = function() {
     EXPERIMENTAL: numerical addressing
   */
 
-  this.getPropertyAddress = function(path) {
+  this.getAddress = function(path) {
     var doc = this.getDocument();
     var nodeId = path[0];
     var property = path[1];
@@ -155,7 +155,7 @@ Container.Prototype = function() {
     return path;
   };
 
-  this.getNextPropertyAddress = function(address) {
+  this.getNextAddress = function(address) {
     var nodeId, node, properties;
     var doc = this.getDocument();
     // extra implementation for the most common case
@@ -168,7 +168,7 @@ Container.Prototype = function() {
       } else {
         if (address[0] < this.nodes.length-1) {
           nodeId = this.nodes[address[0]+1];
-          return [address[0]+1].concat(this._getFirstAddress(doc.get(nodeId)));
+          return [address[0]+1].concat(this.getFirstAddress(doc.get(nodeId)));
         } else {
           return null;
         }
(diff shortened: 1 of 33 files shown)


obuchtala:
@michael could you give me your feedback on the new address based API?
Some examples:
- Getting an address for a path:
  https://github.com/substance/substance/blob/b11916f50a0d0e3b1e90accea1654ad197335c5c/test/unit/document/container.test.js#L12
- Traversing addresses:
  https://github.com/substance/substance/blob/b11916f50a0d0e3b1e90accea1654ad197335c5c/test/unit/document/container.test.js#L49
  https://github.com/substance/substance/blob/b11916f50a0d0e3b1e90accea1654ad197335c5c/test/unit/document/container.test.js#L119
- Getting a range of paths:
  https://github.com/substance/substance/blob/b11916f50a0d0e3b1e90accea1654ad197335c5c/test/unit/document/container.test.js#L194
- Comparing addresses:
  https://github.com/substance/substance/blob/b11916f50a0d0e3b1e90accea1654ad197335c5c/document/container_selection.js#L234
  https://github.com/substance/substance/blob/b11916f50a0d0e3b1e90accea1654ad197335c5c/document/container_selection.js#L243

Some places where this is used already:
https://github.com/substance/substance/blob/b11916f50a0d0e3b1e90accea1654ad197335c5c/document/transformations/copy_selection.js#L77

The good thing about addresses is:
- they are sortable/comparable
- address is an array of indexes which represent childIndexes

From an address you can get to the top-level node using:

```
container.getChildAt(address[0]);
```

The last segment of an address is always the property index. E.g. in paragraphs there is only one property, so this is always 0. In a structured node, such as a figure, there are multiple properties. The propertyName can be revealed using:

```
node.getComponents()[_.last(address)]
```

michael:
Thanks, I will have a look later! Need to go now.

michael:
This is good stuff! I think it's a powerful concept to allow hierarchical addressing (for lists etc.) while still allowing a flat view of components. I read through the tests, and I have not much to add. I'd like to test this inside the writer app, but there are still too many things breaking.

obuchtala:
Hmmm. Maybe I was too fast with removing the container annotation index regarding smooth transition.
Should I recover that, so that we can try this out iteratively in the writer?

michael:
Not sure, probably it's more efficient to keep it a radical change. :) But maybe we should bring it together when I'm back. Also I think this should be tested first with the editor interface, then moving on to the writer. I'd rather continue with my Writer Refactor efforts, while it's getting better, there are still some smelly things I'd like to get rid of.

Maybe you have time to give me some thoughts regarding #63.

obuchtala:
@michael plz review and merge

michael:
Looks very good. I didn't see anything break in the Lens Writer integration. Merging.

PR #45, opened 2015-09-13
OO.extend() refactor

obuchtala:
Improved to make it easier to customize the extend implementation.
Plus: now OO.extend() allows for mixins.

### DIFF
--- a/basics/oo.js
+++ b/basics/oo.js
@@ -18,10 +18,15 @@ var defaultKeyProps = {'name': true, 'displayName': true};
 var _inherit;
 
 var extend = function(parent, keyProps, afterHook, proto) {
+  if (arguments.length > 4) {
+    var args = Array.prototype.slice.apply(arguments, 3);
+    proto = _.extend.apply(null, args);
+  }
   var Constructor = function ExtendedClass() {
     this.__className__ = proto.displayName || proto.name;
     parent.apply(this, arguments);
     if (this.init) {
+      console.log('DEPRECATED: we want to drop this built-in hook in favor of a custom hook.');
       this.init.apply(this, arguments);
     }
   };
--- a/basics/oo.js
+++ b/basics/oo.js
@@ -19,7 +19,7 @@ var _inherit;
 
 var extend = function(parent, keyProps, afterHook, proto) {
   if (arguments.length > 4) {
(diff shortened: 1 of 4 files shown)


michael:
So init goes away in favor of didInitialize?

obuchtala:
No built-in initializer anymore. didInitialize is not built-in but rather called by Component and now also by Node.

michael:
Ah, yeh that's good I think!

michael:
Can you quickly explain:

``` js
var extend = function(parent, keyProps, afterHook, proto) {};
```
- parent is clear.
- keyProps are value properties (no functions?)
- afterHook?
- proto only used for mixin or do I generally have to pass prototype functions there?

obuchtala:
keyProps: Some of the props in the proto object may be keywords, i.e., they may be treated in a different way, such as name goes to constructor.static.name.
afterHook: look at the data node implementation.
proto: as you would expect, plain object, or now also a sequence of plain objects

obuchtala:
Note: you never call this function. You call Node.extend which has everything bound already, and only expects the proto.

michael:
Yeah I'm trying to finally understand this makeExtendFunction thing.

OK so when I now do Foo.extend I can use it like this:

``` js
Foo.extend({
  properties: {
    name: "Foo",
    type: "bar"
  },
  myFunc: function() {}
});
```

where `properties` is now treated specially to store properties on the class, which would be equivalent to:

``` js
Foo.static.name = 'Foo';
```

?

Thanks!

obuchtala:
Not in general, but if Foo is extending Node, then the Node afterHook will put the properties (~schema) into Foo.static.schema. It also puts 'name' into Foo.static.name. This is defined by Data.Node when calling OO.makeExtensible.

obuchtala:
Please look at a Node class... you see properties, name and also displayName in use.

michael:
Ah and they go to the schema then.

https://github.com/substance/substance/blob/oo-extend-refactor/data/node.js#L194-L196

And when I would not use the extend pattern for nodes I would specify this.properties on the Foo.Prototype... like you did in Container.

Good stuff. Hit the merge! :)

obuchtala:
Exactly.

Formerly I had this in a static property, but this was inconsistent with the defineProperties magic. So I dropped that, and moved to defining it on the proto.

PR #47, opened 2015-09-14
OO.extend() and OO.inherit() support static properties.

obuchtala:
- either defined using OO.makeExtensible(clazz, staticProperties)
- or via proto.static = {...}

Look at test (oo.test.js) for usage.

### DIFF
--- a/basics/oo.js
+++ b/basics/oo.js
@@ -13,11 +13,11 @@ var _ = require('./helpers');
  */
 var OO = {};
 
-var defaultKeyProps = {'name': true, 'displayName': true};
+var defaultStaticProps = {'name': true, 'displayName': true};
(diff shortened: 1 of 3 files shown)


obuchtala:
@michael review plz

obuchtala:
Note that proto.static is built-in, as it is an object merged into static as opposed to a single property copied over.

michael:
So

``` js
Foo.extend({
  static: {name: 'bla'}
});
```

now works on any Substance.OO object by default..?

obuchtala:
Yes. Good or bad?

michael:
:+1:

michael:
trying to apply it on my branch...

PR #58, opened 2015-09-18
Writer Refactor

michael:
Some considerations:

I think our current Writer extension mechanism is a bit over-engineered and makes things a bit intransparent for the new user. I wonder if we should consider making a step back, and put state handlers in the custom writer implementation rather than doing them within this external state handler file. I think in most cases Writer apps will be pretty custom and just tie different components together, and you make assumptions on the complete system, like sending a user to the edit citation panel in a certain condition. Often it's hard to implement handleStateChange or handleSelectionChange handlers without knowing the context (e.g. other handlers). I'm really not sure but is there a possibility that we are generally on the wrong track here, when we want to have a transparent programming interface for custom writers?

An extension mechanism would make sense more I think more for concrete implementations. E.g. LensWriter could have it's own extension mechanism.

I may be completely wrong also... I don't want to throw away what we have now... we solve many problems with it when it comes to extensions...  but I have to admit it doesn't feel very intuitive to use.

### DIFF
--- a/document/annotation.js
+++ b/document/annotation.js
@@ -62,15 +62,7 @@ var Annotation = Node.extend({
     }
     var text = doc.get(this.path);
     return text.substring(this.startOffset, this.endOffset);
-  },
-
-  // volatile property necessary to render highlighted annotations differently
-  setActive: function(val) {
-    if (this.active !== val) {
-      this.active = val;
-      this.emit('active', val);
-    }
-  },
+  }
 
 });
 
--- a/document/document.js
+++ b/document/document.js
@@ -277,6 +277,33 @@ Document.Prototype = function() {
     return new ClipboardExporter();
   };
 
+  this.getHighlights = function() {
+    return this._highlights;
+  };
+
+  // Set higlights on a document
+  this.setHighlights = function(highlights) {
+    var oldHighlights = this._highlights;
+
+    if (oldHighlights) {
+      _.each(oldHighlights, function(nodeId) {
+        var node = this.get(nodeId);
+        // Node could in the meanwhile have been deleted
+        if (node) {
+          node.setHighlighted(false);  
+        }
+      }, this);
+    }
+
+    _.each(highlights, function(nodeId) {
+      var node = this.get(nodeId);
+      node.setHighlighted(true);
+    }, this);
+
+    this._highlights = highlights;
+    this.emit('highlights:updated', highlights);
+  };
+
   /**
    * Enable or disable auto-attaching of nodes.
    * When this is enabled (default), a created node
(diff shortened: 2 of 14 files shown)


obuchtala:
About extensions: this is what I am saying all the time. I think extension only make sense for a specific application.

michael:
I will try to remove the extension stuff and see how it feels.

michael:
Oli, can I do this, any negative consequences?

https://github.com/substance/substance/commit/17fc04905a9ae945b6cfa55d0b5febb59e6a2e0f

michael:
Another change I've made: Writer now requires a doc instance on construction, the loading stuff must be addressed by the hosting component. Generally it feels good without the extension stuff.

Pls watch this file as I make changes and let me know if anything does not look like a good API.
https://github.com/substance/demos/blob/topic-writer/topic-writer/topic_writer.js

It's kind of hard to decide which things should go into the writer and what should be custom. 

For instance.. would you make **modals** part of the core writer or leave up to the custom implementation? What about the **status bar**? Not every writer necessarily need it. My demo writer app for instance would just not have it. Next is the **save** logic+tool. Shouldn't this be done by the implementer, rather then having this abstracted using the backend. Somehow I feel a save button should not go into the content toolbar. But instead be one level above. On the other hand it could be useful to have it as a ready-to-use component. E.g. the save tool should only be active if there are unsaved changes. Not sure about this stuff really.... just that there should be some decisions made for an easy to use Writer API.

michael:
Introduced an API for highlights on document. Simplifies the state handler implementation for highlights:
https://github.com/substance/demos/commit/af70f296aa1baa70ced3ea7809e7d26fff61d09e

Please review changes:
https://github.com/substance/substance/commit/704cd9bc8c3d469e6416b69265b79d29e7d7cdfa

michael:
Please review and merge if good.

https://github.com/substance/substance/commit/b33ae47a53e9b30297bff1a46b0e8de634d1b26f

obuchtala:
I think the highlight thing should go into surface land as it is purely visual.

obuchtala:
But I think we should deal with it in a separate effort.

obuchtala:
Hmmm.. And I don't like the squash here. There are also some of my contributions. And particularly, there were some issue fixes.
You can use rebase -i if you want to clean up your commits. But leave mine as they are.

obuchtala:
Was this your last commit before you squashed: 704cd9bc8c3d469e6416b69265b79d29e7d7cdfa?
You can find out the original sha using git reflog.

obuchtala:
I extracted my commits from the former version of this PR into its own PR.
Then I rebased this one onto the new master (removing conflicts).

For the future:
1. don't squash contributions of others.
2. I will never again push to your PR/branch.

obuchtala:
Now :+1:

michael:
Hey thanks! And sorry for the squash. Will take care of it in next PR.

PR #62, opened 2015-09-20
Text property manager

obuchtala:
We want to be able to have more control over what is rendered by TextPropertyComponents. I.e., selection, custom annotations, highlights. TextPropertyManager will be the place where every TextProperty within one Surface registers at.


michael:
:+1: This completely replaces the container anno index and also anno fragmenting code? I like how it simplifies the text property component implementation.

michael:
The demos still work with this.

obuchtala:
Yes. While I think it might need some iterations, together with the numerical-addressing stuff it will remove the need of the container annotation index. ~~Most~~ All container things can be implemented without the old index, only with the addressing thing, which always works, also within tx.
This manager thing is for UI.

PR #65, opened 2015-09-21
Introduce Substance Controller interface

michael:
This replaces SurfaceManager and removes some infrastructure code from Substance apps. Also it introduces a shared event mechanism so UI components and Substance concepts (like tools and commands) can communicate with each other.


michael:
Pls review: https://github.com/substance/substance/pull/65/files

Substance.Editor is not updated yet. But otherwise this is close to be merged. Of course not the last iteration on this.

I think it's nice that we know have one interface, that is available everywhere in UI components but also in Component classes.

This does not suffice everywhere... for instance the ContainerNodeComponent still needs to provide a child context surface for the components rendered inside.

Let me know what you think.

michael:
This is ready to merge. The only bug I spotted is related to the TextPropertyManager. See #66.

PR #91, opened 2015-09-29
Introduce keybindings

michael:


michael:
Introducing keybindings abstraction for Surface. I used mousetrap for interpreting the shortcut definitions, however this does not work with browserify so I had to include it via a script tag. I guess we need our own keyboarder again.

michael:
What i have so far is not very mature.. but I think I now understand a bit better what we need. First of all there are two categories of commands and thus essentially separate keymaps.

a) Surface keybindings (cmd+b, cmd+i etc.) we need to handle them here (https://github.com/substance/substance/blob/key-bindings/ui/surface/surface.js#L201) and bind the handlers to the surface element. E.g. if we wanted to overload alt+. and we don't handle it on the surface level, a "…" character would be inserted.

b) Application level / workspace keybindings, cmd+s for saving or tab for switching between editors or fields.

Open questions: How to deal with ctrl+z/ctrl+shift+z? I mean this works more universal, not surface based. E.g. if you add a char in surface 1 and then another in surface 2, but then focus surface 1 again the undo affects surface 2 and set the selection there. Currently we handle this keybinding on the surface, but maybe this should be done on app level.

obuchtala:
> Introducing keybindings abstraction for Surface. I used mousetrap for interpreting the shortcut definitions, however this does not work with browserify so I had to include it via a script tag. I guess we need our own keyboarder again.

We will want to bring back pieces from old Substance.Commander (which contained a refactored version of mousetrap).

> Open questions: How to deal with ctrl+z/ctrl+shift+z? I mean this works more universal, not surface based

Same as with the other CE supported shortcuts, it is important to bind a handler on surface level.
As you said, and as we know from our first writer implemenetation, undo/redo need application level information, so that we are able to open the correct panel when doing undo/redo. This needs to be thought through again.

michael:
I checked google docs how it behaves over there. They don't have really multiple editing areas but if the editing surface is not active ctrl+z does nothing. This brought me to another thought: Maybe it's not such a good idea to support application-wide undo/redo as this will always need a fairly custom interpretation by the app (which surface to show where etc.). I'm not aware of all consequences but what if we tracked undo/redo operations per surface instead of per document. Like say you have a main editor on the left.. and in the info panel there is the  title + abstract editor. it would all be in the same doc instance but if your focus is in the main editor you will undo changes over there..  if you are in the abstract editor you have an undo list only relevant to that editing scope. and then we would not need to stick surfaceId and app state to a doc change....  maybe.. i don't know...

obuchtala:
We actually have only one document instance. Having multiple undo/redo chains would make life pretty hard - every chain needs to be kept 'up2date' so that you can still apply undos after done something 'somewhere else'.
Also if you want to introduce a new panel you would somehow need to be able to manage a new history.
Compared to GDocs we have a much harder life, as we have panels and they don't.

michael:
Very true. They actually have just one ContainerEditor to take care of.. they don't need the ui to respond to undo.

Phew... i really like how many things we are able to solve already.. so i don't want to give up on those features, but of course they come with a cost of complexity.

However it's our strongest point, that you can build full-fledged apps, not just isolated widgets. Hmm.. having solutions for both I guess would be ideal.

obuchtala:
Let's discuss the undo/redo thing separatedly: #102

obuchtala:
Closing this, as it needs a different approach, e.g. IMO we won't want to use Mousetrap.

obuchtala:
Just leaving the patch here:
[commit-f5045d4.txt](https://github.com/substance/substance/files/344524/commit-f5045d4.txt)

PR #94, opened 2015-09-30
Command tests

michael:


michael:
Please review and merge.

obuchtala:
Looking good :+1: 

Sorry, I am still sick.

michael:
No worries. I think this was safe since it's mostly tests. However I have observed some changed behavior during editing, as I switched to use doc.transaction() instead of surface.transaction(). That way I got rid of the surface dependency for ToggleAnnotation commands. But we will check that together when you are back.

I'm doing a documentation iteration now (literal docs) introducing each class and covering only most important public API methods.

Issue #100, opened 2015-10-01
Deal with edge case where there is no focused Surface


michael:
In controller there's the API `ctrl.getSurface()` which returns the currently active surface. However, we don't have an invariant that says that there is always one surface that has the focus. For instance, after the initial load of the editor no surface is selected. So getSurface would return null. This could be a source of errors for users. It would be nice to be able to rely on a value for getSurface. If there is only one surface registered we could just return that... if there are more.. maybe return the first registered one. Or is that too magic.. Ahh just want to have a clear API for interacting.. and working with a surface is fundamental i think.

michael:
We could require the controller to take a surface instance already in the constructor, that being the defaultSurface. So controller always manages 1..n surfaces. but not 0. Like that we would introduce the concept "primary surface".

obuchtala:
I agree, that there should be an initial surface for sake of consistency.
I don't see that a 'primary surface' concept would add much beyond solving the initial problem, as after that the current surface is determined by user interaction.

michael:
We took care of this.

Issue #99, opened 2015-10-01
Unite Surface + Editor interfaces?


michael:
I documented the controller initialisation: https://github.com/substance/substance/blob/improve-docs/docs/controller.md

However, I have a feeling we should try to slim down the interface a bit. Currently we have to instantiate:
- A Controller (formally surface manager)
- An Editor (ContainerEditor, FormEditor)
- A Surface that takes a controller and an editor via construction.

Maybe we could let ContainerEditor and FormEditor inherit from Surface to make the footprint a bit smaller. I understand that Surface takes care only of selection mapping, and that editors define and handle the rest of the behaviour. But maybe we don't need this separation? I like the name Surface as a general abstraction but somehow i feel it could be more natural do things like:

``` js
var bodyEditor = ctrl.getSurface('bodyEditor');
bodyEditor.setSelection(...)
bodyEditor.getContainerId();
```

I'd also prefer to execute commands like toggleStrong on a concrete surface/editor instance rather than letting the controller delegating to the currently focused one. I mean with keybindings we have to do that of course.. but if someone wanted to trigger one those commands explicitly it feels odd to teach them:
- make sure the bodyEditor is focused
- then do `controller.executeCommand('toggleStrong')`

Better: 

``` js
bodyEditor.executeCommand('toggleStrong');
```

What's your opinion?

obuchtala:
I think after all the refactorings (trafos etc.) the editors are indeed not doing very much anymore,
and it would be ok to merge them into Surface.

michael:
Done.

PR #101, opened 2015-10-02
Use Surface as a more central API

michael:
No magic delegation of focused surface anymore.

### DIFF
--- a/document/node.js
+++ b/document/node.js
@@ -83,7 +83,7 @@ var Node = Data.Node.extend({
           this.getDocument().getEventProxy('path').add([this.id, propertyName], this, this._onPropertyChange.bind(this, propertyName));
         }
       }
-    });
+    }, this);
     Data.Node.prototype.connect.apply(this, arguments);
   },
 
--- a/ui/commands/command.js
+++ b/ui/commands/command.js
@@ -2,31 +2,27 @@
 
 var Substance = require('../../basics');
 
-var Command = function(controller) {
-  this.controller = controller;
+var Command = function(surface) {
+  this.surface = surface;
 };
 
 Command.Prototype = function() {
 
   this.getSurface = function() {
-    return this.controller.getSurface();
+    return this.surface;
   };
 
   this.getSelection = function() {
-    return this.controller.getSelection();
+    return this.surface.getSelection();
   };
 
   // Needed for container annos
   this.getContainerId = function() {
-    return this.controller.getContainerId();
-  };
-
-  this.getController = function() {
-    return this.controller;
+    return this.surface.getContainerId();
   };
 
   this.getDocument = function() {
-    return this.controller.getDocument();
+    return this.surface.getDocument();
   };
 
   this.execute = function() {
--- a/ui/commands/toggle_annotation.js
+++ b/ui/commands/toggle_annotation.js
@@ -36,9 +36,9 @@ ToggleAnnotationCommand.Prototype = function() {
     //   return true;
     // }
 
-    // if (sel.isNull()) {
-    //   return true;
-    // }
+    if (sel.isNull()) {
+      return true;
+    }
 
     var annotationType = this.getAnnotationType();
     var doc = this.getDocument();
--- a/ui/controller.js
+++ b/ui/controller.js
@@ -1,272 +1,230 @@
-'use strict';
+// 'use strict';
 
-var OO = require('../basics/oo');
-var Surface = require('./surface/surface');
-var EventEmitter = require('../basics/event_emitter');
-var _ = require('../basics/helpers');
-var Clipboard = require('./surface/clipboard');
-var Registry = require('../basics/registry');
-var Logger = require ('../basics/logger');
+// var OO = require('../basics/oo');
+// var Surface = require('./surface/surface');
+// var EventEmitter = require('../basics/event_emitter');
+// var _ = require('../basics/helpers');
+// var Clipboard = require('./surface/clipboard');
+// var Registry = require('../basics/registry');
+// var Logger = require ('../basics/logger');
 
-var defaultCommands = require('./commands');
+// var defaultCommands = require('./commands');
 
-var Controller = function(doc, config) {
-  EventEmitter.call(this);
+// var Controller = function(doc, config) {
+//   EventEmitter.call(this);
 
-  if (!doc) throw new Error('Controller requires a Substance document instance');
+//   if (!doc) throw new Error('Controller requires a Substance document instance');
 
-  this.doc = doc;
-  this.config = config;
-  this.surfaces = {};
-  this.focusedSurface = null;
-  this.stack = [];
+//   this.doc = doc;
+//   this.config = config;
+//   this.surfaces = {};
+//   this.focusedSurface = null;
+//   this.stack = [];
 
-  this.logger = new Logger();
+//   this.logger = new Logger();
 
-  // Initialize registries
-  this._initializeComponentRegistry();
-  this._initializeCommandRegistry();
+//   // Initialize registries
+//   this._initializeComponentRegistry();
+//   this._initializeCommandRegistry();
 
-  // Initialize clipboard
-  this.clipboard = new Clipboard(this, doc.getClipboardImporter(), doc.getClipboardExporter());
+//   // Initialize clipboard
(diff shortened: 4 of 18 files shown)


michael:
In search for a more concise API to address #99 and #100, I've started an attempt to use the Surface without a controller (surfaceManager) this gives to the user more control about the editor and it makes clear how tools / surfaces / commands are related together.

I've integrated this into the Lens Writer. Check the empowered-surface branch and you should be able to run/review this.

michael:
Related thoughts: https://github.com/substance/substance/blob/improve-docs/docs/simplified-api.md

However, I moved back to this.context.componentRegistry but commands I assign per surface.

obuchtala:
From  your 'simplified-api' proposal:

> - No confusion anymore about focused surface (each surface is managed by the app directly, and can be focused or unfocused)
> - Each command/tool would now operate on surface level (no longer we need dispatching)

I don't understand how you would like to get rid of multi-surfaces, or dispatching.
I.e. when you have a toolbar and two editors you need something like the surface manager, so that you can 'share' the toolbar.
Notice, the fact that a surface is focused doesn't have much to do with the native DOM focus.

Edit:
You should extend your considered use-case by a separated editable title and abstract. Please elaborate how you would see this in your proposal.

michael:
What was bothering me most is that in the command and tool implementations we had to rely on a surfaceManager thing, with that magic getSurface() function, that gives you the currently focused surface. And this getSurface could return undefined...  I wanted to start an attempt where the surface has a more central role and see how far we could get with that approach. Eliminating the surface manager would mean we would have to design our apps so that we have one toolbar per surface. I think this would actually work for most cases.

Questions are:
- Do we really need surface delegation and shared toolbars?
- Could we have multiple editors within one surface?  
  *this would solve title editor vs. main editor in lens writer. we could then also navigate the cursor from the body nodes into the title without blocking. It's also a problem that with our new cursor/selection rendering the user would see two cursors (one in the title) one in the main editor. having one surface that hosts n editors but there can only be one active selection could be nice... *

I'd try how it feels if i design the writer so that editing areas are more independent from each other. For instance:
- Writer
  - ContentEditor (defines a surface context and inside has two editors: form editor for title/abstract and containereditor for main text. there would be one toolbar for both (texttool would be deactivated)
  - FigureEditor (defines a formeditor with individual figures, + a toolbar with only basic annotations, strong, emphasis, link)
  - TableEditor (modal, formeditor with just one table + table-specific toolbar)

You think that'd be possible or make sense compared to what we have? When i started writing docs an explaining i realized that seting up things is not sooo trivial and that this surface delegation makes it harder to interact with one editor (surface!) instance. i'd like to 'sell' to interested substance users that they can create a surface, and attach a containereditor or formeditor or annotator to a dom element, and then they can start using the surface API (setSelection, focus(), etc.)

Thoughts?

obuchtala:
Having multiple editors in one surface (= one CE) makes things difficult on the editing side. How to prevent the title being deleted, pasted over etc... Not that it isn't possible... but shifting difficulties not resolving them. Example, think how you would prevent certain annotations from being applied on the title.

Making it impossible to create an editor with one toolbar sounds really bad.
I agree that we should try to make it simpler if you have just one surface.

michael:
> Example, think how you would prevent certain annotations from being applied on the title.

I would provide a commands list for each sub-editor. But yeah i agree things are just in a different place then. Just thinking what would feel naturally from an API point of view. Just want to emphasize main task of substance.. which is ... create an editor and then interact with it.

> It's also a problem that with our new cursor/selection rendering the user would see two cursors (one in the title) one in the main editor.

You have an idea how that could be solved? And for cursor moving from title to main editor we would need to define some special behavior.

Mhh. I think it should be possible to construct a Surface without registering it to a SurfaceManager. I mean generally it would be good to be able to access different surface instances from anywhere in the app because surfaces will be created somewhere in the component tree.. with no direct access higher level. so some kind of surface registration would be useful then. maybe more as opt-in than as required infrastructure for a surface to work.

obuchtala:
The cursor problem: please notice again the difference between a native focus/blur and surface focus/blur: a surface gets blurred if another surface gets focused. This means that there is only one surface that can have the cursor. Maybe this is not reflected correctly yet, but that would be the way.

To summarize: my feeling is, that there are simply situations where we want to have multiple surfaces (just for the panels, for instance). In these (and only these) cases we need somthing like a surface manager. That means we need a basic solution to this. OTOH, we want to have more convenience so that single-surface scenarios are easy.

Thus, I still think we need some dispatching. I totally agree, that there should be an responsible surface anytime, so that other pieces don't need to bother.

obuchtala:
Furthermore I recommend, we should take full-fledged editors like Archivist to discuss this kind of infrastructure and consequences of proposed changes, instead of demo editors.

michael:
I think all our tools, commands should not know/bother about the managing context. So they should not ask for the focused surface. Instead they should just be updated from the managing instance, and maybe not have a surface reference at all.

Let's try to solve this lens writer / title editor vs main editor thing it's a good usecase. It would be nice if we could make it accessible with cursor movement, and that we don't show the persisted selections when changing the focus between them. This visualize last-selection thing is tricky because in some cases we want it (e.g. when creating a new entity reference workflow) and sometimes we don't: when switching between title and body editors.

michael:
Let's iterate on Lens Writer for the next weeks. It's a nice test bed.. maybe we bring in comments so we also have container-annos in place.

michael:
Thought a bit about a possible API, considering bringing back the SurfaceManager,
but removing dependency to surfaceManager from tools/commands. Still we have a lot to setup,
but maybe we could implement some sort of Boilerplate Component that Substance Writer and
others could derive from. Here is how it could look like:

#### Top level component (Writer)

``` js
this.surfaceManager = new SurfaceManager(doc);

// Dependency Injection
this.childContext = {
  // Surfaces need to register here
  surfaceManager: this.surfaceManager,

  // Tool components register here. toolManager updates tools
  toolManager: this.surfaceManager.getToolManager(),
  textPropertyManager: this.surfaceManager.getTextPropertyManager(),

  // Provided component registry
  componentRegistry: this.componentRegistry
};
```

### Title Editor

``` js
// Sets up a Surface with a form editor.
this.surface = new FormEditorSurface({
  name: 'title',
  commands: [ToggleStrongCommand, ToggleLinkCommand]  
});

// didMount
this.surface.attach(this.el);
this.surfaceManager.registerSurface(this.surface);


// willUnmount
this.surface.detach();
this.surfaceManager.unregisterSurface(this.surface);
```

#### Main Editor

``` js
this.surface = new ContainerEditor({
  name: 'main',
  containerId: 'main',
  commands: [ToggleStrongCommand, ToggleLinkCommand],
  keybindings: ...
});

this.surface = this.surfaceManager.createContainerEditor({
  name: 'mainEditor',
  commands: [ToggleStrongCommand, ToggleLinkCommand]
});
```

#### Table Editor (Modal dialog  => surface comes and goes)

``` js
this.surface = new FormEditor({
  name: 'tableEditor',
  commands: [AddTableCell, JoinTableCell, ....]  
});
```

## Thoughts
- We can not get rid of SurfaceManager because we want a shared toolbar for multiple surfaces (writer title+body)
- Tools are now registered to a ToolManager which triggers update calls,
  so tools no longer need to know about the surfaceManager and we can also cover cases where not a selection change updates 
  the tools but a document change (e.g. in a realtime scenario somebody makes a bold annotation where your cursor is)
- easy interaction with surfaces
  - local context: this.surface.executeCommand('selectAll');
  - top level context: this.surfaceManager.getSurface('main').executeCommand('focus');

## Still unsolved
- How to run undo+redo across multiple surfaces?
  - there is surface.executeCommand('undo') but this change could actually affect a different surface
  - should we provide commands on app level, what component should take care of this?
    - surfaceManager.executeCommand('undo') seems weird?!
    - also surfaceManager.executeCommand('save')
    - introduce some kind of documentManager interface for things that are doc-specific not surface-specific?
      - documentManager.executeCommand('undo') would feel better
      - documentManager would then again have a toolManager inside but this time taking care of only document
        tools (like save, undo, redo)
      - feels like a lot of infrastructure overhead
      - maybe better have one interface that manages both things

michael:
We will not do this. That's the new way: #131

PR #103, opened 2015-10-05
Incremental container node updates

obuchtala:


obuchtala:
@michael plz review and merge.

PR #113, opened 2015-10-06
Getting rid of global jquery

obuchtala:
~~We need an indirection to be able to use different jquery implementations (jquery, cheerio).~~
Without any browserify configuration, cheerio will be bundled for the web, even if not used.

Edit: this is now without indirection.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,7 +1,6 @@
 {
   "esnext": true,
   "node": true,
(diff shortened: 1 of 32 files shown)


obuchtala:
@michael let's discuss this

michael:
this works. i rebased controller-refactor to it.

Issue #112, opened 2015-10-06
Problem with changing the selection in a (natively) blurred Surface


michael:
Reproduce (in Lens Writer):
- click somewhere inside the toolbar to blur the surface (selection gets rendered as a fragment)
- click into the text again

Expected:
- Cursor should move to the new position

Actual:
- Cursor jumps to the previously position

PR #117, opened 2015-10-07
Fixing #112

obuchtala:


obuchtala:
@michael can u try this out?

obuchtala:
Unfortunately, these kind of things can not be tested, as it isn't possible to emulate pointed clicks (as opposed to element clicks).

obuchtala:
This also seems to resolve that strange scrolling issue.

PR #129, opened 2015-10-09
Introducing Jsdoc

obuchtala:
There are some things to be considered so that jsdoc creates the desired output:
- in every 'sub-module' (e.g. 'basics') we put a `@module` declaration into the `index.js` file: https://github.com/substance/substance/commit/1b0afa73e0334672b1ac153f744ac1b1dbab3be3#diff-23333edbbf1e9d87c7d79217bd9930d4L12
- other modules within the main modules (e.g., basics/oo) should be declared using '/', e.g. `Basics/Helpers`: https://github.com/substance/substance/commit/1b0afa73e0334672b1ac153f744ac1b1dbab3be3#diff-c3f5846a2605df71445fde36a4c5b03eR10
  There is no support for child-modules in jsdoc, so they will not show up within the main modules documentation. However we can document them as properties there: https://github.com/substance/substance/commit/1b0afa73e0334672b1ac153f744ac1b1dbab3be3#diff-c614e68c2023701920b90f24c358a3a9R18
- classes should have a `@class` annotation. To let them show-up within the modules documentation we need to bind them using `@memberof`: https://github.com/substance/substance/commit/1b0afa73e0334672b1ac153f744ac1b1dbab3be3#diff-764613c7e9e3b43466d7382faf0d4f9aR10
- class prototype functions need to be bound to the class using `@memberof`: https://github.com/substance/substance/commit/1b0afa73e0334672b1ac153f744ac1b1dbab3be3#diff-e57a1df2675e7c92dbac37fb85fc5bf3R24
- static class functions should be bound with: https://github.com/substance/substance/commit/1b0afa73e0334672b1ac153f744ac1b1dbab3be3#diff-c67a8a8345dc0f893fce1a0f0c571281R121

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,5 +28,5 @@ node_modules
 .lock-wscript
 
 dist
-doc
+dist/doc
 test/tmp
--- a/.jsdoc
+++ b/.jsdoc
@@ -1,0 +1,3 @@
+{
+    "plugins": ["plugins/markdown"]
+}
--- a/basics/error.js
+++ b/basics/error.js
@@ -7,8 +7,8 @@ var OO = require('./oo');
  *
  * @class SubstanceError
  * @extends Error
- * @constructor
- * @module Basics
+ *
+ * @memberof module:Basics
  */
 function SubstanceError() {
   Error.apply(this, arguments);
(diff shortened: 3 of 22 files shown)


obuchtala:
Good enough for a first iteration.

PR #131, opened 2015-10-10
Surface and controller as component

michael:
Surface and Controller are now regular components. We are more opinionated now, but it makes the setup of an editor much easier.

I adapted `surface.test.js` to the new interface, however I wonder if we could simplify the boilerplate a bit. Maybe you have ideas. 

I will further improve this work, but wanted to get this into master fast. Please review and merge.


obuchtala:
Looking good to me.

PR #133, opened 2015-10-10
Enable api doc generation for 'document' module.

obuchtala:


obuchtala:
Just testing the api doc hook.

PR #136, opened 2015-10-13
UI polish

michael:
Mainly documentation, but also merged editor interface into Surface interface for ContainerEditor/FormEditor.


michael:
For testing this you can use substance/lens-writer#edge or substance/demos#bleeding-edge

michael:
Oh and for the demos, only the ProseEditor example works yet. However it's a nice example to see the new API for making a custom editor.

Here's a getting started example on the new API, it's a custom ProseEditor component, that mainly consists of configuration.

https://github.com/substance/demos/blob/bleeding-edge/prose-editor/app.js

It got easier in my opinion since you just extend controller and build your editor inside that scope. No longer you have to define childcontexts etc. if you don't want to provide custom ones.

obuchtala:
Looking good to me :+1: Editor still works :)
Configuration in app.js also pretty readable.

PR #192, opened 2015-11-06
Documentation Generator

michael:


obuchtala:
You might want to have a list of properties and methods in a class node, right?

obuchtala:
And you might want to have an 'Object' type for classical modules.

michael:
Definitely. Gimme a sec :)

obuchtala:
Maybe we want to add a static flag to properties and functions, too...

michael:
should we make properties separate nodes with a class PropertyNode ? Or we just provide an object here:

``` js
[{name: 'doc', type: 'model/Document', description: 'The document instance'}]
```

Same for params, props in Components etc.

obuchtala:
I think we should keep everything flat. I.e., a property/function has an id which is added to classes.
Dunno...

michael:
just thinking everything that can not live on it's own could be provided as a rich object on the 'parent' node if you want so. but then at the same time.... it's nice to have the model+component pairs everywhere.

I have it a bit mixed now.. Method is a separate node type where properties are provided as an array of objects, all stored on the class. What is preferrable?

obuchtala:
Hmmmm... substancish would be if everything would be addressable on its own.

E.g.,

var Component = {
  type: 'class',
  id: 'ui/Component',
  properties: ['ui/Component#el']
}

I am not sure yet... but nesting rich objects seems to be against our common substance way...

obuchtala:
So... referring to your last q, I think method should be separated...

michael:
we could still have it all flat in our body container making scopes implicitly using id's

``` js
[
{id: '/ui'} // namespace
{id: '/ui/Component'} // class
{id: '/ui/Component-Description'} // text thingee
{id: '/ui/Component.prototype.append'} // method
{id: '/ui/Component.prototype.append~element'} // param
]
```

not sure...

obuchtala:
Hmmm.. If we want to be able to exploit references (e.g., addressing in links) we need to be able to resolve those targets, which could be 'ui/Component#el' for instance.

obuchtala:
I.e., every existing entity should be accessible via that uuid.

michael:
This kind of namepath (jsDoc) equivivalent will be important yes. Ideally from the path=id you can already read what category something is.

obuchtala:
exactly :)
:+1:

michael:
What path for a method param then for instance?

obuchtala:
Hui... maybe addressing a param goes too far

obuchtala:
?

obuchtala:
I think, addressible should be, everything which can be required and Class members (static or instance).

michael:
think so too.. but then params don't have to be separate objects in general.

``` js
myFunc.params = [{name: 'doc', type: 'model/Document', description: 'A Substance document instance'}]
```

obuchtala:
agree

michael:
I like using # for `instance` but we have a conflict between methods and properties here.

Component#el vs Component#render

obuchtala:
This would be impl detail then... still we could introduce extra nodes for params later... but with random uuids.

obuchtala:
the same in js in general

obuchtala:
I.e. in javascript there can't be prop and function for the same name... right?

obuchtala:
So I think we just need to distinguish static vs instance

michael:
true.. so we could go with # for both.. only drawback would be that you can't tell from the path if it's a method or property.

obuchtala:
we have indexes for that... and also get(id).type

michael:
but then you are just one click away ;)

obuchtala:
fine :)

michael:
To address a property of an object: `model/documentHelpers.getAllAnnotations` ?

obuchtala:
:+1:

michael:
ok.. good for today :) see you in the morning!

obuchtala:
sleep well

michael:
So do you!

michael:
It's looking so good! Will give the styles some more attention tomorrow.

We should integrate our api introduction markdown file. Currently at: https://github.com/substance/substance/blob/master/.jsdoc/index.md

Integral:
Could you send a link?

michael:
You can check out substance-doc branch of substance. Then `npm install` and then `npm start`.

Docs are ad-hoc generated at: http://localhost:4201/doc

We need figure out some more stuff, but we will get there soon. It's really nice we have control over everything now.

obuchtala:
We should consider to separate general doc generator sources from the substance specific things.
I.e., making the doc generator part of API.

obuchtala:
@michael Regarding my last comment: we should treat ComponentComp et al. as package :)

michael:
Let's try.  But let's not enter doc generator business. ;)

obuchtala:
Merging this as a first iteration.

PR #234, opened 2015-11-13
Component docs

obuchtala:


obuchtala:
I want merge now.
Things are getting too complicated with all the active branches, depending on each other.

@michael fine?

PR #236, opened 2015-11-14
Constructors

obuchtala:
Plus: 
- MemberContainerComponent parent for all member containers (ns, module, class)
- refactored MemberIndexComponent and used them in all member containers


obuchtala:
@michael check this out :)

obuchtala:
and merge if it is working

michael:
How nice! Checking...

michael:
Will try to tweak a bit and merge when ready. 

I want the regular constructor to show up as new Classname(param1, [param]). I also want to implement a specialised ComponentConstructor node that shows as $$(Toolbar, props) and below a parameter (=property list) by just delegating to the regular ParamsComponent.

How should we make the switch between the custom substance-specific component constructor and the regular one? Where to store the custom information in the model? Should we provide a custom ClassNode implementation that has an isComponent flag or introduce an extra ComponentNode + ComponentComponent? 

We are almost there now! :)

obuchtala:
This dispatching is not yet possible.
I want to pass through custom tags, adding an object property to DocumentedNode.

You could go ahead and think how we could address this if there was an tag property on every DocumentedNode:

```
tags:
  <name>: {
    content: 'string'
  }
```

michael:
yeah i think that way we keep the model extensible without touching code. On the view side I would just provide a custom ConstructorComponent that looks into those tags and renders a or b.

Isn't his information already available? Just stored as an array?

obuchtala:
Nope... That is what I meant with passing through...
Am 14.11.2015 1:02 nachm. schrieb "Michael Aufreiter" <
notifications@github.com>:

michael:
ok. Then i wait for that.

Issue #235, opened 2015-11-13
Unified Element API


obuchtala:
From #226 evolved another element API, which is now a bit confusing.

We decided introducing an abstract DOM Element API which is used to abstract jquery/cheerio away and which we would use for Component's VirualElements, and Component, too.

michael:
Yeh. Best thing about that abstraction is that we can define and document one interface for Element manipulation, used in Component's render methods as well as in converters. 

Regarding the name: Calling it just Element would be too generic mh?

obuchtala:
Maybe too generic, yep... Maybe you want to sleep over it...

obuchtala:
I think it is a DOMNode actually (could also be text or comment)

michael:
Ya i mean somehow the relation to the DOM is there.. For Substance.Components it would be just the DOMElement facet of it...  while there is more comonent-specific.. like setProps setState etc.

obuchtala:
Yes... would you add setProps to the generic interface? don't know... it seems wrong to me.

obuchtala:
Hmmm.. it could also be in a dom namespace... and then it could be dom.Node / dom.Element ?

michael:
hmm what else would go into dom then? It's clearly also ui.. so maybe ui/dom/Element but I guess ui/DOMElement would not look worse.

obuchtala:
ok. ui/DOMElement is fine with me.

obuchtala:
@michael can u help me with naming?
When I have `ui/DOMElement` being the abstract interface, how should I call the implementation which is using jQuery/native DOM?

obuchtala:
In the other cases it is clear: `ui/VirtualDOMElement`, and Component itself.

obuchtala:
~~`BrowserDOMElement`?~~ nah, it is working for cheerio elements as well

Edit: so for clarification... this implementation works in the browser but also server-side. It uses jQuery/cheerio for many things, and some things are implemented using the respective native implementation.

obuchtala:
An alternative from the Java world: calling the interface `ui/IDOMElement` and the implementation `ui/DOMElement`

obuchtala:
We also could hide the implementation from the programmer, in the way that they would never require the implementation.
E.g., in an importer or an exporter, they would use `el.$$` or `el.createElement()`
In Component renderers it is ok to have it the way it is currently, i.e. using `Component.$$`.

[...]

PR #242, opened 2015-11-14
Converter refactor

michael:


michael:
I've done an iteration on the demos (in future examples), covering the new converter API:

It's looking good, the API looks really clean and we have a clear separation now between Article definition, converter the editing app world.

@oliver---- can you review:

https://github.com/substance/demos/tree/beta2/note
https://github.com/substance/demos/tree/beta2/converter
https://github.com/substance/demos/tree/beta2/notepad

From my side we can merge the converter refactor. I will adjust also the Lens converter, but this could also be done after the Beta 2 release. 

#235 we may want to address before though?

Edit: #235 is done

obuchtala:
@michael Thx I am just starting the next iteration

obuchtala:
We can address #235 in an extra PR.

Examples looking good so far.

obuchtala:
We should wait until we have merged #234.

obuchtala:
I will make this PR independent of #234.

obuchtala:
@michael don't push to this branch for now
I am rebasing it onto master and creating a new PR

obuchtala:
I finish the exporter things before we can merge.

obuchtala:
Need to fix ClipboardImporter and ClipboardExporter too

obuchtala:
So! first iteration done.
Still I am using jQuery internally, but it has gone from the interface.
Later on I might try to get rid of it completely.

@michael u can find a test for HTMLExporter to see how the exporting spirit is.
Plz let me know if you think this could be improved.
And merge if you are happy so far.

obuchtala:
I will address other converter related stuff, such as Clipboard in another PR.

michael:
No rush on eliminating jQuery as long as we don't use it directly. I will utilize the exporter API in the note converter example now.

packages/embed/EmbedHtmlConverter.js
```diff
@@ -0,0 +1,21 @@
...
+    node.src = el.attr('src');
+    node.html = el.html();
+  },
+
+  export: function(node, el) {
+    el.attr('src', node.src)
+      .html(node.html);
+  }
+
+};
```
michael, line 21: A minor thing, but do if we want to be strict that filenames of classes start with a capital letter, we should call it `embedHtmlConverter.js` since we return an object here, not a full class. Or is that too picky?

michael:
Worked just fine: https://github.com/substance/demos/commit/cd60ee8378b16286c356c56852a0cd8cb3f993ff

michael:
Want to upgrade the Lens Importer so we can check how this plays with XML. We want to have XmlImporter XmlExporter entry points right? However we should get this merged now and work on the polishing in separate PR's.

packages/embed/EmbedHtmlConverter.js
obuchtala, line 21: I agree.

obuchtala:
Did we go for XML now in the science blogger?

obuchtala:
Beyond that, having XML converters for all of our nodes would be fine.

michael:
Lens Writer is already XML yes. But abusing the old HtmlImporter API for importing that.

For most nodes XML converter would be the same as HTML importer (e.g. paragraph). should we introduce redundancy there? Maybe we could call the generic converters just DOMConverters?

obuchtala:
I would rather add a delegator. Then there is no confusion.

obuchtala:
@michael I would need a lens version working with latest master. Is it ok if you address that LensConverter update next?

michael:
Will do that tomorrow!

obuchtala:
Cool :)

PR #245, opened 2015-11-15
Custom tags

obuchtala:
- passing through custom tags
- detecting type description in custom tags

### DIFF
--- a/doc/components/DocumentationTocPanel.js
+++ b/doc/components/DocumentationTocPanel.js
@@ -4,6 +4,8 @@ var oo = require('../../util/oo');
 var TocPanel = require('../../ui/TocPanel');
 var Component = require('../../ui/Component');
 var $$ = Component.$$;
+var MemberContainerComponent = require('./MemberContainerComponent');
+var NamespaceComponent = require('./NamespaceComponent');
 
 function DocumentationTocPanel() {
   TocPanel.apply(this, arguments);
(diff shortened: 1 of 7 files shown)


obuchtala:
@michael 

I was thinking about how to document components.
Instead of introducing a constructor for the $$ API it seems more straight-forward to add '@props' and '@state' tags to the '@class' and render that with a special ComponentComponent.

obuchtala:
merge when u r fine with it.

obuchtala:
Merging.. we will continue in a new PR.

michael:
Let me think... That approach sounds reasonable... just see the following problems:
- constructor would not show up as a member in the memberindex
- constructor as a class member is addressable through deep linking - we would not have that functionality for components then (bit unconsistent, also nice to be able to link to the constructor when there's a longer description for the class + examples.

What if we had the info stored as you describe but don't touch the ClassComponent at all but introduce a custom constructor component that checks if it's parent is a component by inspecting its tags? and then render either the $$ thing if it detects a component or delegate to the regular ConstructorComponent?

obuchtala:
The problem is, there is no such constructor thing. Even hard to give it a consistent name.
What would you then show in the member index?

Edit: simply said, there is actually no constructor.

michael:
No i mean.. we document the constructor as usual. In a separate comment block with @param tags and the class itself just gets the @component flag. Then we would have a constructor node also for components. and can show it in the member index?

obuchtala:
How would you render it in member index?

obuchtala:
From a technical point of view: Components have a private constructor and an external so called 'named constructor' which is a static function of Component. (which makes it more a factory method than a named ctor) 
Edit: not even a factory. Component.$$ just creates a ComponentElement. So there is actually no way to create a Component directly.

michael:
Hmm true.. mah maybe we should not document them at all. Just via example?

obuchtala:
I would prefer to have a nice dedicated section for props and states in a component's documentation.
Most often you don't document the rest of a component's API, just what it is and how to use it.
You could go for a manually crafted doc, though, reusing the type notation is convenient.

michael:
Agreed. It's more like a usage section in the Component case.. not a constructor in a typical sense.

Something like this structure:

**TextPropertyEditor**
Component defined in ui/Component.js#...

The text property editor.

**Props:**
- path Array path to text property
- commands SurfaceCommand[] array of SurfaceCommand classes

**Example:**

``` js
$$(TextPropertyEditor, {
  name: 'authorNameEditor',
  path: ['author_1', 'name'],
  commands: [EmphasisCommand]
})
```

obuchtala:
:+1:  And maybe even a states section (in other cases).

michael:
That's how we do it! I will work on this in the evening. I think I have everything I need data-wise?

obuchtala:
yup :)

PR #254, opened 2015-11-16
Fixed TocPanel and moved member categorization logic into model.

michael:


obuchtala:
The highlighting of the active item could be visually stronger.

obuchtala:
Seems to work. I merge.

obuchtala:
Old problem: when using the TOC items, it scrolls alright, but then highlights the item above.

obuchtala:
Leaving this open due to #256, #255

obuchtala:
#256 was false alarm. #255 can be fixed later.

PR #258, opened 2015-11-17
Refactoring Node API.

obuchtala:


michael:
Crazy! :) This will be not backward compatible right? So for Lens even if you'd finish this soon, I will base the converter on the old API?

obuchtala:
No.. it is breaking.

obuchtala:
As I see now, it is only the nodes itself and some core stuff. So it won't be too much work to bring it in shape.

obuchtala:
Continueing the discussion from https://gist.github.com/oliver----/c0ae4b9e09aaf67efde3

michael:
Cnt. from gist: What happens currently when we have a getter setup for this.content that reaches out to this.props.content, but we do that thing I described:

```
  function Paragraph() {
    DocumentNode.apply(this, arguments);
    this.content = 'blabla'; // is the getter already in place here and would complain?
  }
```

obuchtala:
that thing you described?

michael:
setting this.content in constructor.

obuchtala:
It would just delegate to this.props.content.
You can not distinguish who is actually calling `this.content`.

michael:
ahh sorry makes sense of course.

obuchtala:
~~Do u agree, that this is too critical, and so we stick to having a 'props' scope (-> similar as in components actually)?~~

Edit: Hmmm.. actually the critical part, detecting such collisions we can just not solve.

obuchtala:
Should we switch to using `node.props.content`. But it is just so inconvenient :(

Otherwise, as we can't avoid collisions, we could also store values directly. It just doesn't make any difference.

michael:
yeah we can leave it.. I see that we need some kind of namespacing. In Component it's a bit different though since we never access props via getters/setters, so there's no magic there.

comp.props.containerId rather than comp.containerId. Not sure though if that's an option for Node. It would also break a lot of existing code.

obuchtala:
I think the major difference to Component is, that component props are used from within only:

```
this.props.content
```

but nodes from outside

```
node.content
```

Which makes the explicit scope unattractive for nodes.

obuchtala:
So my lesson learned is that our property getters don't give us anything, and we can just drop them.
Potential for collisions is the same, so we don't change nothing with doing that move.

obuchtala:
Summarizing the schema definition thing. I don't like the incremental nature of the `addProperty` as you suggested.
Instead we could

```
Property.static.defineSchema(
...
);
```

obuchtala:
Using `new Property()` or a declarative description is then just matter of taste, i.e. both need some interpretation implementation, which would be pretty identical.

obuchtala:
Similar as with selections I would prefer to write it in a declarative way (no need for extra requires) and maybe create such Property instances internally.

```
Paragraph.static.defineSchema(
 { name: 'id', type: 'string', mandatory: true, readonly: true },
 { name: 'content', type: 'string', 'default' : '' }
);
```

Then we can fail early.

obuchtala:
Another ninja move ;)
To get rid of the extra 'components' field we had, we could use a specific type for text properties.

```
Paragraph.static.defineSchema(
 { name: 'id', type: 'string', mandatory: true, readonly: true },
 { name: 'content', type: 'text', 'default' : '' }
);
```

Then we would know which properties are meant to be changed via Editor.

michael:
Not sure i understand wether you want to drop property getters for nodes now or not? ;)

The only bad thing of using type names is that it's more error prone then passing a class. E.g. we may want to change a type's name and then we need to find all positions. i mean it would also happen if the filename changes but then we would fail earlier with the bad require.

I mean after all i want to say that the current node system did not cause a lot of headache even if not properly type checked.. so we are on a good path anyway. i'm not having strong opinion for one solution or another. it's hard for me to compare them atm, considering all consequences.

michael:
The order could change in the view for text 'components' though, which would then need to be reflected in the defineSchema block?

``` js
Paragraph.static.defineSchema(
 { name: 'id', type: 'string', mandatory: true, readonly: true },
 { name: 'content', type: 'text', 'default' : '' }
);
```

obuchtala:
What I want to say:
1. property getters are obsolete, as they do not solve the collision problem.
2. for types we would need to interpret and check validity, in both approaches. We could indeed use explicit types via providing a class, to communicate such a strong relationship. Both approaches are failing early.

obuchtala:
> The order could change in the view for text 'components' though, which would then need to be reflected in the defineSchema block?

The same problem as currently. As long we do not derive the order from the view this is not solvable. OTOH, this is actually pretty difficult, when not even impossible: rendering is done very late, and for manipulation (e.g. transforms) we need that information much earlier.

obuchtala:
Do you have an idea to approach that problem?

I.e. for container manipulation and container selections we need to be able to derive the actual structure.
E.g., you create a ContainerSelection spanning a paragraph, and reaching into a figure, say the whole title plus a part of the caption. Then you would need to know that structure during the deleteSelection transform. So there is a dependency of the model to the actual rendering in the view.

michael:
What I don't like about the declarative approach is that we have a lot of implicit interpretation of those type names even for primitives. Why is it 'string' and not 'String'? Do we support all native javascript types? At least we should document it somewhere.

Somehow for the native types something like this would look nicer and be more explicit. And we don't have to do interpretation on type names.

``` js
MyNode.static.schema = {
  id: String,
  level: Number,
  tags: [String],
  bibItem: BibItem // this needs a require which could be a bit annoying, but in return it's very explicit
};
```

obuchtala:
Yes we can do that. Totally agree.

Still this is a declarative description of the schema :p

michael:
On the other hand Javascript itself works a lot with string type identifiers.

typeof 24 -> 'number'
typeof 'foo' -> 'string'

so we would not get too far with it. But i really don't know...

btw.. why is `'asdf' instanceof String`  === false?

obuchtala:
I think you are just talking about how to specify types. I am fine with using classes.
Still I don't want to require Property all the time.. that is where I am insisting on a declarative approach... not for the types.

michael:
Regarding your previous question: I think we should not allow those strange selections and instead have a NodeSelection which selects the whole structured node. It can either be selected completely or not. No half-overlap stuff. Would make our lives easier and be a usability improvement in my opinion.

obuchtala:
Hmmm... internally the only difference between a figure and a paragraph is that the figure has multiple text properties.

obuchtala:
> btw.. why is 'asdf' instanceof String === false?

dunno

obuchtala:
With the container stuff... still I need a way to derive the order, e.g., for container annotation highlighting. Or do want to prohibit this too?

The key question is, how to interpret a ContainerSelection.

michael:
Not sure i understand. Can you give an example?

obuchtala:
Give me an answer to this: which properties in the document are addressed via 

```
ContainerSelection({ container: 'main', startPath: ['p1', 'content'], startOffset: 0, endPath: ['figure1', 'caption'], endOffset: 10})
```

As you will see you can't tell much without knowing more about properties and their layout.

michael:
I see i mean.. for the figure example i don't think it's important to allow container annos span into figure nodes or even over a figure node.. i think i would prevent this.. even so you can't wrap a figure in an container annotation.

however there could be other types that just have two text properties and you want to allow container annos and then you need the order. i'm really fine with encoding the textproperty order somehow. just thinking of a good format.

in general i think we should tend to make complex structural nodes more like atomic pieces. and not try to support the text-optimized concepts. u agree?

obuchtala:
> I see i mean.. for the figure example i don't think it's important to allow container annos span into figure nodes or even over a figure node.. i think i would prevent this.. even so you can't wrap a figure in an container annotation.

I think it should be possible to select a paragraph and a figure together.

> however there could be other types that just have two text properties and you want to allow container annos and then you need the order. i'm really fine with encoding the textproperty order somehow. just thinking of a good format.

Any suggestion?

> in general i think we should tend to make complex structural nodes more like atomic pieces. and not try to support the text-optimized concepts. u agree?

Maybe. In general I am thinking to make the ContainerEditor 'region' aware. So that it behaves like a PropertyEditor in structural nodes. But dunno yet how to exactly achieve that.

obuchtala:
Still I like my original suggestion to use a specific type for text properties.

michael:
So for the format... I don't like the {name: 'id'... } stuff.. having object keys feels more natural for schemas. since the textorder is so special to our editing i would not change the schema spec format for it.. rather have a separate place to describe it (as it is now)

``` js
MyNode.schema = {
  properties: {
    title: String,
    caption: 
  },
  defaults: {
    content: 'foo'
  },
  textPropertyOrder: ['title', 'caption']
};
```

mah also quite verbose.. need to think more.

obuchtala:
Also we 'agreed' upon having a method to define the schema, so that we have early failing.

michael:
> I think it should be possible to select a paragraph and a figure together.

yup! But the figure as a whole then. I think that's how word also works. (do they have dedicated fig captions actually?)

michael:
defineSchema method is good yeah. Sorry mixing up things as i experiment with syntax ;)

michael:
away for lunch for a bit...

michael:
sorry i was wrong.. word has images more like a big character.. i don't like it actually.

obuchtala:
```
var Text = require('substance/model/DocumentSchema').Text

MyFigure.static.defineSchema({
  title: { type: Text, default: "foo" },
  imgSrc: { type: String, default: "http://" },
  caption: { type: Text, default: "bar" }
});
```

Without defaults:

```
MyFigure.static.defineSchema({
  title: Text,
  imgSrc: String,
  caption: Text, 
});
```

And maybe `Text` comes with `default=""` already.

obuchtala:
> > I think it should be possible to select a paragraph and a figure together.
> 
> yup! But the figure as a whole then. I think that's how word also works. (do they have dedicated fig captions actually?)

This means we might need a different concept for ContainerSelections/ContainerAnnotations ?

obuchtala:
From what we discussed about container related stuff so far, a short intermediate summary:

We only need the order of properties for structural nodes. Simple nodes like paragraph have only one text property (still we would need to know which one), and Hierarchical nodes are taking care of dynamically.
For structural nodes, a way could be, not to support deep ContainerSelections which select only parts.

If that all works, from a schema point of view we still need a way to designate the one text property for simple nodes (note: there might be other string properties in the schema).

Structural nodes would be editable only in a FormEditor way, or each property on its own.
It would be great to teach the ContainerEditor _regions_, so that it behaves accordingly in a structural node.

obuchtala:
```
MyFigure.static.defineSchema({
  title: "text",
  imgSrc: "string",
  caption: "text", 
});
```

```
MyFigure.static.defineSchema({
  title: { type: "text", default: "" },
  imgSrc: "string",
  caption: "text", 
});
```

obuchtala:
Damnit. When following the above approach:

```
MyNode.static.defineSchema({...});
```

we don't have the context where to put the schema.

I.e., we would need an extra argument:

```
MyNode.static.defineSchema(MyNode, {...});
```

so that we can attach the schema to the actual Node class.

@michael ur 5 cents?

michael:
Hmm.. either that ye.. or we go back to later interpretatin (when the node gets constructed) and just use `MyNode.static.schema = {...}`.

obuchtala:
SHAKA!
I just tried out in node4. in static methods, `this` is bound to the class!

:)

michael:
what to do until then? :)

obuchtala:
In our static thing we can do what we want.. such as storing the class into the static scope.

obuchtala:
This is exciting :)

obuchtala:
I tend to drop the oo.extends API totally?
@michael  u fine with it?

michael:
Yes. Maybe deprecate it first.

obuchtala:
Yep.. will do that.

obuchtala:
@michael  are you fine with the naming of these schema type props?

```
id: { type: 'string', required: true, readonly: true }
```

obuchtala:
As an alternative to `required: true` we could make this default, and break out using `optional: true`.

Edit: and having `default` set is implicitly optional.

michael:
yeah why not.. i mean.. i don't know what i prefer in this case really.

michael:
Supposedly there are less optional props than required ones. ;)

obuchtala:
Agreed :)

obuchtala:
Closing this as I moved to a different branch.

PR #259, opened 2015-11-17
Make Importer XML-aware.

michael:


michael:
Test suite runs in browser but not on console. You have idea?

obuchtala:
seems to be a phantomjs issue.

obuchtala:
Regarding the contract of parseHtml... despite of implementation difficulties, is this really what you want?

I would expect to get an exact representation of what I am providing as HTML, also allowing for partials.

obuchtala:
- Full specified doc:

```
<html>...</html>
```

parseHtml should provide me with one DocumentElement.
- Single element:

```
<p>...</p>
```

parseHTML should return an DOMElement with tagName p
- annotated text:

```
foo <strong>bar</strong>
```

Should return `[TextNode, StrongElement]`?

The caller needs to know what he is doing anyways. But at least, the API should not be in the way (e.g. by getting a full doc for just a snippet of annotated text).

obuchtala:
While I see that always returning a full DOM document even for only a small partial
  is consistent regarding the return value, it is actually not consistent from a semantical point of view.

  I.e. the result would be the same for 

```
  <html><head>...</head><body>some <strong>annotated</strong> text</body></html>
```

  and for just `some <strong>annotated</strong> text`.

michael:
That sounds good to me. Maybe you can give me some pointers on how i could implement that. I would need to reason about wether we have the full doc context or the scoped context. I would want this same semantics also for the parseXML then. For example when i get some XML how do I detect based on the markup that i want a full doc vs that i want just an element.

Example input:

``` js
// considered a full doc
<article><body>hello</body><article>
```

vs. 

```
/// considered a single partial element
<p><span>foo</span></p>
```

Since any elements are allowed in xml we can not reason about the semantics of <article> etc.

obuchtala:
Hmmm.. true. I need to try out the XML parsing first.

michael:
Regarding the problem with the phantom.js testsuite.. it seems that inBrowser evaluates to false and then jQuery is used (with it's strange parsing semantics). Shouldn't there be a window object in phantomjs?

Also: Could we try to avoid jQuery for the parsing? I mean when on the server we can do cheerio, when on the client or phantomJS we have DOMParser which is the preferable option.

obuchtala:
DOMParser in phantomjs seems to parse null... :(

obuchtala:
I have an implementation for HTML working on all 'three' platforms.

obuchtala:
Now I will play with parseXML and see what we have here...

michael:
Nice!

Reading through the code. You have solution regarding the broken DOMParser in phantomjs?

obuchtala:
Working on a phantomjs solution... jQuery $() is not good for parsing whole docs...

You would get ArticleEl not [ArticleEl]. if number of root els > 1 then you get [].
You can look into the DOMElement.test.js to see how I want it to be used.

michael:
Ah I overlooked that.. makes sense. Yeah jQuery $ for parsing full HTML is bad as it cuts away the top level elems...  I mean we will never have a problem here except in PhantomJS.. is there a way to disable only those three tests when run in phantomjs?

obuchtala:
I am trying to go the XML way in phantomjs...

michael:
you mean $.parseXML  for full html docs?

obuchtala:
right.

michael:
Somehow what is a strange is that we test against a different target when tests are run automatically. Better than nothing of course but for instance if something goes wrong in the DOMParser branch of the implementation we would only realize it when we run the browser test suite?

However, it wouldn't be bad to have the whole thing tested with jQuery anyways.

obuchtala:
Yeah. We should take a look at options to start a browser.
I mean with karma we were able to do that. It is just so heavy weight for only spawning a browser and get a result... Need to think about that.

obuchtala:
Can we merge this?

michael:
:+1:  maybe squash the commits a bit? there's a lot of noise.

obuchtala:
Will do

michael:
before you do.. please kill  this.convertDocument function from HTMLImporter.

michael:
or make it abstract.

obuchtala:
I mean, as a reference implementation it makes sense actually. Or doesn't it?

michael:
Ok then this is it.

``` js
  this.convertDocument = function(elements) {
    this.convertContainer(elements, this.config.containerId);
  };
```

I just still did not generalize the Importer to be aware of multiple containers. Will add this as an issue for beta 3.

obuchtala:
Finally I made it abstract and added documentation instead.

michael:
bumped lens#dev accordingly.

PR #264, opened 2015-11-19
Human-readable names for text-type properties through SwitchTextTypeTool

michael:
We abuse the SwitchTextTypeTool a bit here, since this functionality is not directly related to text type switching. However it's a perfect place to show that information. One 'flaw' is that I had to activate the SwitchTextTypeCommand for all property editors (titleEditor, abstractEditor). I think it does the job for now, but we should seek for a more consistent solution eventually.


michael:
Don't merge. I forgot some edge cases.

michael:
Now we are good I think.

obuchtala:
I think it is fine as a quick-fix.
We will try to find a proper solution to it in future.

Just as quick estimation: probably we want to update the label independently from the dropdown.
I.e., the label could be part of a different package, which SwitchTextType would depend on.
So that we don't need to activate the switch tool to show the label. But don't know yet how to do that.

obuchtala:
I will try it out first before merging.

obuchtala:
We need to iterate on this one. It was the first time I could check now and it seems broken.

I.e., while it shows the context now, it doesn't let me switch type elsewhere.

obuchtala:
@michael attention, I have rebase to latest master. So do a reset to FETCH_HEAD.

obuchtala:
Well, I have reviewed it yesterday. It didn't work for me. Should this have changed? don't see commits...

obuchtala:
ahh sorry you mean the refinement...

michael:
I solved the categorization stuff here in this branch.

https://github.com/substance/substance/commit/48eec04805255af06eec4d5d6c175df8f565b393

obuchtala:
Works.. also switch text type.
U can merge.

PR #267, opened 2015-11-19
Refactored node api

obuchtala:


obuchtala:
I am working more on core refactorings. Just having this PR to document the first iteration.

There are some 'questionable' things like the 'text' property type. I.e.,

``` js
MyNode.static.defineSchema({
  content: 'text'
});
```

gets compiled into

``` js
MyNode.static.defineSchema({
  content: { type: 'string', addressable: true, default: '' }
});
```

I.e., 'text' is just for convenience, and maybe to obscure my internal representation.
@michael what is your feeling about that? Do you have a better name for 'addressable'.
Why I have chosen this name is, that it is used only by `model/Container#getAddress` etc. to navigate through editable properties.
The strategy to obscure that thing is of course questionable, though, it opens me the chance to change that mechanism without touching all the node schemas.

obuchtala:
I will merge this into the core-refactorings branch.

PR #268, opened 2015-11-19
Refactor class extend.

obuchtala:
As suggested by @jacwright I generalized Class.extend to also be able to do the work of oo.inherit

### DIFF
--- a/model/AbstractDocument.js
+++ b/model/AbstractDocument.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var oo = require('../util/oo');
 var each = require('lodash/collection/each');
 var EventEmitter = require('../util/EventEmitter');
 var IncrementalData = require('./data/IncrementalData');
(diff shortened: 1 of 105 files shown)


obuchtala:
@michael another API twist with Class.extend
Can you have a look how you like using Class.extend as a replacement for oo.inherit
https://github.com/substance/substance/blob/aaf8dadcdda02877bd8a0086168d755410633c0d/test/unit/util/oo.test.js#L59

This would mean we don't need to require `oo` all over the place.

obuchtala:
At the same time, the way we have done by now, is still working.

obuchtala:
@jacwright thx for the inspiration.
I integrated the idea, still with our protoype based static scope implementation, and supporting legacy.

obuchtala:
@michael see how I can drop oo imports here https://github.com/substance/substance/commit/7d3fa00a2bc06fd38f0ad55799fe88b315e29adf

obuchtala:
Also, now mixins are much easier to understand.

obuchtala:
Merging.

PR #282, opened 2015-11-24
gulp: generate documentation after bundling

Integral:
pls merge


michael:
This has nothing to do with the autogen though?

Integral:
It is, actually it fixing that error which you mentioned. The problem was that documentation were generated before creation of folder. However there are number of other heroku-related issues. I already close to finish...

PR #285, opened 2015-11-24
Polish

michael:


obuchtala:
@michael I saw that u use the mixin pattern in a way I did not consider.

```
DocumentedNode.extend(ClassNode, MemberContainerMixin)
```

is that actually working?

michael:
Seems to be... What's the correct usage?

obuchtala:
Not that it is wrong... Just didn't think of the case of having a mixin + the Prototype.
But when it's working it is probably fine :)

Edit: i.e. Prototype implicitly taken.

obuchtala:
The explicit way would be

```
DocumentedNode.extend(ClassNode, MemberContainerMixin, ClassNode.Prototype)
```

obuchtala:
But no need to change this, if you find this to be working.

obuchtala:
Just that I need to document it and maybe add a test-case.

michael:
Ok :) need to go now. But you can review and merge. I tested with notepad
example, not with lens yet.

obuchtala:
Will do :)

obuchtala:
reviewed :+1:

obuchtala:
Will merge when :frog:

Issue #286, opened 2015-11-24
Upgrade dependencies


michael:
Bump dependencies to latest versions. E.g. lodash, cheerio, jquery but also dev dependencies and check that everything runs smoothly still.

I think we should specify explicit versions in package json (no >= stuff) and bump manually from time to time.. just so users don't have bad surprises.

@oliver---- agree? @Integral want to take that one over?

obuchtala:
:+1:

Integral:
Thank you!

Integral:
ok
@michael what should i do here in terms of verefication? run tests and check examples?

michael:
Yep! tests, examples, and substance/lens#dev should still run. Most importantly we need to test fresh npm install if everything goes write. We will all try once you are ready.

Integral:
Ok!

Integral:
@michael give the next one while you are testing!

PR #289, opened 2015-11-25
bump dependencies to latest versions

Integral:
issue #286
test are ok
todo example is ok
lens is ok (references and images are not working but it's not related i think, at least no errors)


michael:
:+1:

michael:
Tested and works fine for me too.

PR #291, opened 2015-11-27
Selection docs

Integral:
Updated container and property selections.


Integral:
Hey there,
i just realized that maybe I shouldn't touch some docs, as it's not so clear what is public API and what is not... For example look at 083454b90802b6a9a14da286e5e18e48c6e40d21. @oliver----?

obuchtala:
No worries. You can document them. I will go through it tomorrow and add `@private` to those we don't want show. :)

PR #304, opened 2015-11-28
Apply surface focus workaround only in FF.

### LINKED ISSUE #301: surface.el.focus() causes a surface scroll when the surface is currently blurred
Reproduce in Lens:
- select some text
- click somewhere into the toolbar to natively blur the surface
- toggle an annotation

This triggers a Surface#setSelection which finally calls focus(), which triggers the unwanted surface scroll. 

https://github.com/substance/substance/blob/64a75988e207b03a6e2375d98ffbbac79a42d4c7/ui/Surface.js#L680-L687

Since this call to focus is only needed for FF maybe we should guard that so it doesn't get called for Chrome. I will now check if the scroll issue also exists for FF.

obuchtala:
Fixes #301.

### DIFF
--- a/ui/Surface.js
+++ b/ui/Surface.js
@@ -53,10 +53,13 @@ function Surface() {
   this.domObserverConfig = { subtree: true, characterData: true };
   this.skipNextObservation = false;
 
-  // set when editing is enabled
-  this.enabled = true;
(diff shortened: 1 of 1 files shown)


michael:
:+1:

PR #313, opened 2015-11-30
Implementing Component 'owners' correctly.

obuchtala:
Before, owners where the last Component instance entered during rendering,
and not the Component triggering the rendering.

### DIFF
--- a/doc/app.js
+++ b/doc/app.js
@@ -1,7 +1,6 @@
 var Documentation = require('./model/Documentation');
 var Component = require('../ui/Component');
 var DocumentationReader = require('./DocumentationReader');
-var $$ = Component.$$;
 var $ = require('../util/jquery');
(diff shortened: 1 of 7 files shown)


obuchtala:
@michael I pushed fixes to substance/demos#new-mount-api  and substance/lens#new-mount-api to make them work with this.

Edit: didn't change package.json, as I am using npm link ATM. Make sure you are testing with the right substance version.

michael:
Looks like we miss to preserve elements in some cases. E.g. in lens#new-mount-api, when a citation is selected, the main section gets wiped and rerendered. Thus the scroll pos jumps to the beginning of the article.

obuchtala:
I could extract the observed issue into a test case. This should be ok now.

obuchtala:
Closing in favor of #346

PR #318, opened 2015-12-03
Correct filename case

asmecher:
Looks to me like there's an upper/lowercase mismatch between TOCPanel and TocPanel. This PR renames the file (not sure whether you'd prefer that or renaming references to it).


obuchtala:
Thx. That is exactly how we want to do it.
I will merge this after making sure that we require it correctly in other places.

michael:
Fixed this for Lens. Merging. Thanks @asmecher.

asmecher:
Good sharing a bottle of wine with you!

obuchtala:
Yes indeed! That was nice :)

Issue #325, opened 2015-12-04
Better browserify errors


michael:
We currently use this code for our ad-hoc browserify bundling:

``` js
app.get('/app.js', function (req, res, next) {
  // var startTime = Date.now();
  browserify({ debug: true, cache: false })
    .add(path.join(__dirname, "app", "app.js"))
    .bundle()
    .on('error', function(err){
      console.error(err.message);
      res.send('console.log("'+err.message+'");');
    })
    .pipe(res);
});
```

However, this gives us crappy errors e.g. when a require path is wrong:

```
Cannot find module 'substance/ui/ScrollPane' from '/Users/michael/projects/substance/lens'
```

It only tells us the folder where the problem happened but not the actual file.

@Integral maybe there is a way to find out what the actual file was that did the bad require. Maybe you have some time to investigate this.

Integral:
@michael i think it is not possible without source maps...

Integral:
and i'm not sure that we want to introduce source maps...

obuchtala:
I think we already have source maps (`debug: true`).
Just the browserify error is unspecific. Maybe there is a solution out there?

obuchtala:
Not solving this issue, but just as a reference what others do about errors.

http://latviancoder.com/story/error-handling-browserify-gulp
https://gist.github.com/apfelbox/b2d5ee9fe32e5b42adb2

Integral:
i saw this, but it's about gulp...

obuchtala:
I fear, that is not browserify's fault:

https://github.com/substack/node-resolve/blob/5cae82fb22cb64d5b72f703c787dc0fd418ed412/lib/sync.js#L17-L33

The implementation uses dirname for the error :(

Integral:
To confirm your fears I could say that in other cases we have normal error trace...

obuchtala:
There is a PR: https://github.com/substack/node-resolve/pull/64

obuchtala:
Maybe it gets fixed soon.

Integral:
Khm... It's there from beginning of year

obuchtala:
For now we could fork browserify and patch together a version with this https://github.com/jmm/node-resolve/commit/4eb424ba952af0c651fcaf14e05395b146c541dc included

obuchtala:
Hmmm.. yes, true :(

Integral:
Do we really need this if it's related only to `Cannot find module` errors?

obuchtala:
[...]

PR #326, opened 2015-12-04
use browserify fork to get more info on errors

Integral:
Fixing #325


michael:
How nice! I will use this in the various projects (Lens, Stencila etc.)

Issue #327, opened 2015-12-04
Put dev server stuff into a serverUtils module


michael:
Can we put our dev server stuff into a lib? So we don't have to write it all over again?

``` js
app.get('/docs/app.js', function (req, res) {
  browserify({ debug: true, cache: false })
    .add(path.join(__dirname, 'doc', 'app.js'))
    .bundle()
    .on('error', function(err){
      console.error(err.message);
      res.send('console.log("'+err.message+'");');
    })
    .pipe(res);
});

var handleError = function(err, res) {
  console.error(err);
  res.status(400).json(err);
};

var renderSass = function(cb) {
  sass.render({
    file: path.join(__dirname, 'doc', 'app.scss'),
    sourceMap: true,
    outFile: 'app.css',
  }, cb);
};


app.get('/docs/app.css', function(req, res) {
  renderSass(function(err, result) {
    if (err) return handleError(err, res);
    res.set('Content-Type', 'text/css');
    res.send(result.css);
  });
});

app.get('/docs/app.css.map', function(req, res) {
  renderSass(function(err, result) {
    if (err) return handleError(err, res);
    res.set('Content-Type', 'text/css');
    res.send(result.map);
  });
});
```

@Integral you want to take that?

Please find a good API for it... maybe provide a middleware thing, so it can be used like this:

``` js
app.use('/app.js', serverUtils.bundleJavascript(path.join(__dirname, 'app', 'app.js')));
app.use('/app.css',serverUtils.bundleStyles(path.join(__dirname, 'app', 'app.scss')));
```

Integral:
Yep.

Integral:
I did try to create more general thing, like router which you can connect with line of code, but it looks like in express you can't send any params to router. And we need it for path etc. So here what we have #328.

Integral:
BTW, do we want to use koa instead of express?

michael:
Done.

PR #328, opened 2015-12-05
use serverUtils module for dev stuff

Integral:
for #327

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,12 +2,12 @@ var express = require('express');
 var path = require('path');
 var glob = require('glob');
 var browserify = require('browserify');
-var sass = require('node-sass');
 var PORT = process.env.PORT || 4201;
 var app = express();
 
 var config = require('./doc/config.json');
 var generate = require('./doc/generator/generate');
+var serverUtils = require('./util/serverUtils');
 
 // use static server
 app.use('/docs', express.static(path.join(__dirname, 'doc/assets')));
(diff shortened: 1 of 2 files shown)


michael:
Looks good. I think this can go into the main substance repo though.

Integral:
Hm... Where I should put it?

michael:
util/server.js ?

michael:
Could you make it so that sourcemaps are generated by default?

So that this:

``` js
app.use('/docs/app.css', serverUtils.bundleStyles(path.join(__dirname, 'doc', 'app.scss'), {file: 'app.css'}));
app.use('/docs/app.css.map', serverUtils.bundleStyles(path.join(__dirname, 'app', 'app.scss'), {file: 'app.css', map: true}));
```

becomes just one line?

Integral:
If we are doing this in the same repo, than we could do it even in the one
line for js/css.

Integral:
But once you want do it from external app, it is not possible. So this will not work say with lens.

Integral:
So let's decide...

michael:
Can you give examples? I'm not sure I follow for the two options we have.

Integral:
@michael  there is no option if you want to use it in other applications
if you want to use one middleware for a few routes (app.css, and app.css.map), then you actually need express Router instance, but you can't send any parameters to this class. So you there is no way to tell what is the path to Router.

What we could try to do is to play with paths like '/docs/app.css*', so it will match both app.css and app.css.map, then we could use real middleware and pass any data there. But I'm not sure that it is possible...

michael:
Still not sure I understand completely. Do you think something like that would work?

``` js
app.use('/docs', serverUtils.serveApp({
  js: path.join(__dirname, 'doc', 'app.js'),
  css: path.join(__dirname, 'doc', 'app.scss')
}));
```

If app.use does not work we can also do it like this:

``` js
serverUtils.serveApp(app, {
  js: path.join(__dirname, 'doc', 'app.js'),
  css: path.join(__dirname, 'doc', 'app.scss')
});
```

I really don't know the best practices in Express land. Regarding koa.. I would say not now. We use only so little things and Express is definitely familiar to more people.

netsgnut:
I think use of `express.Router` to separate modules is currently one of the common practices found in the express ecosystem.

Chiming in for this another pattern, also commonly found:

``` js
app.use('/docs', serverUtils.serveBundle(function (bundle) {
  bundle.add('app.js', path.join(__dirname, 'doc', 'app.js'));
  bundle.add('app.css', path.join(__dirname, 'doc', 'app.css'), { map: true });
}));
```

What `serveBundle` does is just like what @michael's first sample code does, and returns an [express.Router](http://expressjs.com/en/guide/routing.html#express-router).

Integral:
serverUtils.serveApp(app, {some data}) is good suggestion, will try

Integral:
@michael check latest commit please

michael:
Thanks a lot. Will review this tomorrow.

michael:
I did some consolidation and added docs. I found out that you can serve the CSS source maps inline also, so we don't need two routes for it.

Integral:
ok

PR #349, opened 2015-12-18
Handle triple click events to fix selection behavior.

michael:
Inline nodes (contenteditable=false islands) lead to unexpected
behavior on triple click (selection gets blocked). This fix ensures the
expected full text property selection is made.

### DIFF
--- a/ui/TextPropertyComponent.js
+++ b/ui/TextPropertyComponent.js
@@ -27,6 +27,7 @@ function TextPropertyComponent() {
 
 TextPropertyComponent.Prototype = function() {
 
+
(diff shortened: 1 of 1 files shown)


michael:
We may want to capture the event one level higher at the wrapping block element. Then you could also triple click next to the text, as well as between the lines.

obuchtala:
I need to rethink the whole thing, as also the Edge double clicking stuff is revealing a problem with our current approach in Surface.

obuchtala:
Implementing this thing in Surface works fine.
Will close this PR and provide mine.

Issue #50, opened 2015-09-17
Booting with an Empty Editor Content 


netsgnut:
Tested on 1.0.0-beta. If we are to [replicate the Notepad demo but with an empty content state, i.e. empty `<div id="editor_content"></div>`](https://github.com/substance/demos/blob/bcfcd6919a769ed4a1d46e3e28d1960849bc2c57/notepad/index.html#L16-L21), we would result in an cryptic `TypeError: node1 is undefined` bubbling from [here](https://github.com/substance/substance/blob/fed41a88a6ffb051bbf27a3e8afa8d1deab5bafe/surface/surface_selection.js#L31) once the editor is getting focus.

The obvious workaround is, of course, make sure there is at least _something_ at the editor holder, but in some cases starting from scratch sounds reasonable. Any ideas, or am I simply misusing it? Thanks.

michael:
Editor should be robust regarding empty content. Will be fixed for Beta 2! Thanks for reporting.

netsgnut:
Thanks for the assurance! :)

michael:
So things changed a bit here.. the Notepad example requires a doc instance as a prop, so conversion is now a task of the component user. However we should check how the editor deals with an empty container. Will come up with a solution for that.

obuchtala:
I think ContainerEditor should always make sure that there is at least one node to interact with.

obuchtala:
We discussed that this should be done UX wise. I.e. the ContainerEditor should a message and provide a button or such that would create a first node.

michael:
Fixed via #361.

PR #361, opened 2015-12-20
Deal with empty containers.

michael:
Attempting to solve #50.


michael:
Please review using substance/examples#master and commenting out all content in notepad/index.html

https://github.com/substance/examples/blob/master/notepad/index.html#L13-L17

I've seen some problems with selection mapping on empty containers. Maybe you can cover those cases. Let me know when I have your ok on this and I will do another style iteration and then merge.

obuchtala:
Solved this by not attaching event handlers when container editor is empty.

obuchtala:
Now you should be able to do the finishing.

michael:
We have too many branches. I merge this into master and we can hot fix if there are problems.

Issue #360, opened 2015-12-19
Comparing container addresses does not work correctly


obuchtala:
At the moment, container addresses are just plain arrays. At several places we do comparisons like

```
address1 < adress2
```

which essentially takes the string representation for comparison.
This is obviously wrong: as `'[21,1]' < '[3,0]'`, for instance.

Instead we should use a dedicated Address class with :

```
Address.prototype.isBefore = function(other) {...}
Address.prototype.isAfter = function(other) {...}
Address.prototype.equals = function(other) {...}
```

Issue #149, opened 2015-10-17
Inserting nodes with no components breaks


michael:
Reproduce:

``` js
      var img = tx.create({
        id: Substance.uuid("image"),
        type: "image",
        src: 'path/to/image.png',
        previewSrc: 'path/to/image.png'
      });

      return surface.insertNode(tx, { selection: args.selection, node: img });
```

Image does not have any editable areas (=components) but this is assumed in Container._handleInsert.

It fails here because `_getNodeComponents` returns an empty array and causes `_insertComponentsAt` to crash.

![image](https://cloud.githubusercontent.com/assets/2931/10560079/266bbc3c-7500-11e5-9e7b-e0858f5c1cf4.png)

michael:
I think we should not require a DocumentNode to have components. Seeing this assumption in various places. E.g. node.js. 

``` js
  getComponents: function() {
    var componentNames = this.constructor.static.components || [];
    if (componentNames.length === 0) {
      console.warn('Contract: a node must define its editable properties.', this.constructor.static.name);
    }
    return componentNames;
  },
```

Let's support non-editable nodes which have no `Node.static.components` value set.

michael:
Node.statc.components -> Node.statc.fields   (editableFields)

michael:
Maybe we can get rid of explicit component listing by deriving that info again from the view component. Since we no longer depend on React, we can analyze our component's VirtualDOM.

michael:
I assume this is no longer an issue... will check.

michael:
This is still not addressed in the current implementation. It fails here:

![image](https://cloud.githubusercontent.com/assets/2931/11918138/226e4f6c-a725-11e5-8e72-5cb271def3f1.png)

getAddressablePropertyNames returns an empty array here. And of course you can not determine a path for a non-existent property.

What semantics should getPathForAddress have for nodes that have no text properties? Should it just return null? How we make code that depends on getPathForAddress robust?

michael:
I tested this but I still see the same error. Can we investigate that together?

obuchtala:
I am working on this. Needs to be solved in Container.

obuchtala:
Now this should be working basically. Not perfect yet, but ok to continue.

michael:
That's been fixed.

PR #365, opened 2015-12-22
Dedicated class for container addresses

obuchtala:
Fixing issue #360


obuchtala:
This is ready to merge when green.

obuchtala:
Fixing #149

michael:
Will get this into master now, even though the image deletion is broken now. We can fix that in another PR.

PR #366, opened 2015-12-22
Integrating of highlights into TextPropertyManager.

obuchtala:
…


obuchtala:
Obsolete in this form.

PR #378, opened 2015-12-28
Surface: getDomNodeForId was using this.element instead of this.el

khlbrg:
getDomNodeForId is now getting element through method getElement


khlbrg:
I'm not sure you guys want me/us to create PR rather then just creating issues?
But anyhow, this time I'm doing a PR :)

michael:
PR is perfect. Thanks! :)

PR #382, opened 2016-01-08
#381TableCellIterator throws OutOfBounds error on `while(iterator.next()) {}` 

dmukhg:
I will be pushing a patch fixing the test case within a couple of hours.


obuchtala:
Thx!

PR #385, opened 2016-01-11
Fix #384: Use findAll('tr') rather than find('tr') to get a list of all table rows

### LINKED ISSUE #384: TableHTMLConverter tries to iterate over a single tr node (uses find() not findAll())
As seen in the [link](https://github.com/substance/substance/blob/d214d786650c967a9affb78dfde14512950dc141/packages/table/TableHTMLConverter.js#L35), the result of el.find('tr') is iterated over. This needs to be replaced with el.findAll('tr') which actually returns an array. el.find('tr') returns a DefaultDOMElement containing the first element that matched 'tr'.

dmukhg:


obuchtala:
Thx!

PR #391, opened 2016-01-13
Test XMLExporter

michael:


michael:
Having troubles fixing the broken Clipboard tests. 

I fixed this one: https://github.com/substance/substance/commit/f274b55d1283370620346eb8b068242d11b3a97d

Here we have the situation that we have root, a DefaultDOMElement that wraps the document node. So we can not call createElement on it, as this requires ownerDocument to be set.

I wonder if we should allow wrapping documents into elements at all? If we allow it, then `rootEl.createElement` would feel natural though and we should make it work somehow.

Then in ClipboardImporter we have some occurences of `this.$$` which still points to the static `DefaultDOMElement.createElement`. Now the question is, should we play the same game for importers and store a this.state.documentEl (either an empty XML doc or HTML doc?). Our first assumption was that we only need this stuff for the exporter.

PR #431, opened 2016-01-31
[Fix #430] TableSectionHTMLConverter uses obsolete `el.tagName = ...` syntax

### LINKED ISSUE #430: TableSectionHTMLConverter uses obsolete `el.tagName = ...` syntax
This method no longer exists in DefaultDOMElement because the tag name is considered immutable. This should be changed to `el = el.withTagName`.... syntax.

dmukhg:
We are reassigning a function parameter. DOMExporter.convertNode
requires us to return the element.


michael:
Thank you!

PR #438, opened 2016-02-03
Provide context to Commands on construction.

michael:


obuchtala:
Looking good.
Rebasing this onto master.

PR #441, opened 2016-02-04
Surface refactor

obuchtala:
- DocumentSession is now real owner of selection.
- Surface is taking over what TextPropertyManager did.
- Custom PathAdapter is gone. I am using lodash get/set/setWith now.
- lot of document change analysis is now done in DocumentChange.

Switched to lodash 4.2.1
Switched to cheerio 0.20.0


obuchtala:
Very likely we gonna see regressions. But it was definitely worth it. ;)

PR #443, opened 2016-02-04
Resizable split pane

michael:

PR #446, opened 2016-02-07
Collaborative editing

michael:


michael:
First early version is ready.
- currently each single local change gets sent to the server `['commit', [c1], localVersion)]`
- commit gets confirmed by the hub `['commitDone', newServerVersion]`
- other collabs get notfied with an update `['update', [c1], newServerVersion]`
- No handling of concurrent changes yet.

When I type a char on left editor, the right editor gets the selection set after the update has been applied.

@oliver---- can we do something about that? If a passive editor (has no native selection) gets a remote change it should not render a dom selection. We would not have that 'problem' in an iframe szenario. So maybe that's another way to go next.

@oliver---- maybe you have time to have a quick look. You can test using the collabwriter example (substance/examples#collab). I will in the meanwhile work a bit on #443.

michael:
It seems I'm not correctly applying the changes still. The following gives wrong results:
- In left editor make a word strong
- In left editor type a char inside the strong anno (text op + anno op is created -> correct)
- Now in the right editor type a char inside the strong anno (only text op is created, so anno does not see an update)

obuchtala:
For sure we have some more work to do in Surface etc.
I will take a look at your implementation and think about next steps...

obuchtala:
@michael I think as a first iteration we can merge this into master now.

PR #460, opened 2016-02-14
More modular CollabSession.

michael:
Does no longer require a websocket immediately. An explicit open call now connects the CollabSession to the remote endpoint.

We need to update the collabwriter example to conform the new API.

See substance/notepad for usage and how websocket reconnecting is handled: https://github.com/substance/notepad/pull/9/files


michael:
Closing in favor of #472.

PR #461, opened 2016-02-14
Fixing issue 458

obuchtala:
There was a regression.
But also some weirdness: Range.setStart(startNode, startOffset) + setEnd(endNode, endOffset) does not behave the same as Range.setStart() + window.getSelection().extend(endNode, endoffset)
I now do the selection.extend() for all non-collapsed selections.


obuchtala:
There is still a rendering issue when the ContainerSelection reaches into the next paragraph at char 0.
It is a Fragmenter bug.

PR #472, opened 2016-02-20
Improved collabsession

michael:


michael:
The CollabWriter example is working again. Only thing missing is the test suite. I don't know how to provide the same message queue dump. Messages look differently now because we changed the message format to an object notation.

obuchtala:
Will do an iteration.

michael:
I used a branch `collab-refactor` in the examples. Let me just check quickly if my yesterdays changes did not break anything for the examples.

michael:
Looking good. Feel free to merge into `collab`.

michael:
Ready for merge with master?

obuchtala:
@michael I think u want to clean up first: https://github.com/substance/substance/blob/a28845db4ec2697f24a9ec43e0f13bfc78259c9b/collab/CollabSession.js#L49-L103

obuchtala:
But then it is ready to get merged.

michael:
Oh. True. Will do!

michael:
Working on including snapshot API into CollabHub too. Maybe i can that in as well today. What's your opinion about the current file structure? Didn't have a better idea than moving collab-related features into a separate collab folder.

obuchtala:
I think that's ok, as it is not in the way. We should add a dedicated README later.

michael:
Sounds good. I will merge when green.

PR #486, opened 2016-02-24
Collab: pass userId to backend when storing changes

Integral:


michael:
This breaks the testsuite. You would need to update TestBackend also to make the test suite pass again.

michael:
TestStore it's called. You could also rename to TestBackend and make changes in the CollabHub.test.js file.

Integral:
@michael there are browserify or lodash issues...

```
[19:31:29] Starting 'test:karma'...
24 02 2016 19:31:29.879:WARN [plugin]: Error during loading "/Users/danielbeilinson/Projects/substance/node_modules/karma-browserify" plugin:
  Cannot find module 'lodash/collection/reduce'
24 02 2016 19:31:29.890:WARN [preprocess]: Can not load "browserify", it is not registered!
  Perhaps you are missing some plugin?
[19:31:29] 'test:karma' errored after 326 ms
[19:31:29] Error: No provider for "framework:browserify"! (Resolving: framework:browserify)
```

michael:
Mhh karma depends on old lodash, but that should work fine when dependencies are resolved correctly. I triggered restart of the test run.. sometimes it has a hickup.

Integral:
I also can't pass tests locally. Because of this.

obuchtala:
I had this too yesterday, which resolved after wiping node_modules and npm installing again.
It was strange, as it took like two times. Seemed to be a node 4 problem or even an issue with using nvm.
So happy wiping ;)

Integral:
Say it to travis-ci ;)

obuchtala:
Travis might also just be choking... I restarted that one failing core.

obuchtala:
But... hearing that both of u are not able to run tests locally makes me whine ;)

obuchtala:
@michael  @Integral plz let me know when u have resolved the local testing issues.

michael:
Tests were running fine for me locally.

michael:
@Integral i need to finish a bunch of stuff, but will look at your PR's and merge them later today. Just one question: Are all the PR's independent or does one need the other to be applied first?

obuchtala:
Ok then there is no prob.
Travis is choking sometimes. So whenever you see it failing out of nothing then just try restarting the build.

Integral:
@michael, they are independent in repo context. But both has siblings in notepad repo.
@oliver---- Works good after wiping. Before wiping I saw that karma-browserify haven't got lodash inside its's node_modules folder (but have dependency in package.json). Strange npm...

PR #487, opened 2016-02-24
File uploading

Integral:
It should work. At least in modern browsers.
Note that there is something wrong with saving of figures. It's not saving and not broadcasting, but without errors.
Hubclient will emit uploading progress percents.

### DIFF
--- a/collab/HubClient.js
+++ b/collab/HubClient.js
@@ -198,25 +198,31 @@ HubClient.Prototype = function() {
     Takes a file object and returns an url (location of the uploaded file)
   */
   this.uploadFile = function(file, cb) {
-    var formData = new window.FormData();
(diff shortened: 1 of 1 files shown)


michael:
> Note that there is something wrong with saving of figures. It's not saving and not broadcasting, but without errors.

Don't understand. First you say it works, then it doesn't work ;) What means broadcasting? Sorry, perfect confusion ;)

Integral:
Uploading is works and you will see the image. But other collaborators will not see it as well as you after refreshing the page.

michael:
Ah ok. But that's on my end then.

michael:
Tested and works. For progress I will do a separate iteration, but does not have priority yet.

PR #501, opened 2016-03-03
Use signatures instead of separated arguments.

Integral:
@michael could you take a look?


michael:
Looks good. Merging.

michael:
Maybe in the tests you want to use terminology `args` and `result` also, instead of `request` and `response`. It's not wrong but i would use that terminology for the Application Layer, so this could lead to confusion.

PR #508, opened 2016-03-11
Isolated nodes

michael:


obuchtala:
Rebased onto devel.
Need to close and reopen.

PR #538, opened 2016-04-08
Component refactor

obuchtala:
Rewrite of rendering implementation of Component.js
- $$ as parameter for render(): fixes issues with Substance used with two instances of the library code (~peerDependency)
- improved technical architecture
- better debuggability (via setting `RenderEngine.DEBUG = true`)
- automatically retain ancestors of ref'd components
- deep rerendering by default


obuchtala:
Need to fix scrollTo

obuchtala:
Also I did some profiling and it is clear that I need to to optimize some stuff in RenderingEngine.

obuchtala:
Resolved performance problems. However there are some issues with retaining components in delicate nesting scenarios.

obuchtala:
It
is
done

obuchtala:
Still we should wait with merging this, as I have removed routing support, which was not satisfactory.
We have discussed a better approach which I want to try out on top of this branch.

michael:
Nice! And yes, we should bring back the router and make it independent of the component state properties.

PR #540, opened 2016-04-10
Add label to Figure schema.

jure:
If it's missing, the JSON export will miss it too, so when you try to reload the document in the editor, it will fail as label will be undefined here: https://github.com/substance/substance/blob/master/packages/figure/FigureComponent.js#L35

Which will then raise here, as it will try to call _isRawHTML on undefined:
https://github.com/substance/substance/blob/918878959ef88c20b2750963e20fb5750df60f13/ui/Component.js#L1174


michael:
Thanks. We will soon get figure out of Substance and have it a Lens specific node.

PR #541, opened 2016-04-11
Rewriting Routing

obuchtala:


obuchtala:
Added routing to two apps:
- DocumentationReader:
  - router: https://github.com/substance/substance/blob/routing/doc/DocumentationRouter.js
  - initial route to state mapping: https://github.com/substance/substance/blob/routing/doc/DocumentationController.js#L36
  - route updates: https://github.com/substance/substance/blob/routing/doc/DocumentationController.js#L62
- Notes App:
  - router: https://github.com/substance/notes/blob/component-refactor/client/NotesRouter.js
  - initial route to state mapping: https://github.com/substance/notes/blob/component-refactor/client/Notes.js#L100
  - route updates: https://github.com/substance/notes/blob/component-refactor/client/Notes.js#L254-L289

obuchtala:
I am convinced that a no-tech approach is sufficient for a while, maybe forever.
1. Component API is convenient enough to do interesting stuff (such as `comp.find('.my-component')`)
2. Tangling component implementation with routing is poor, as it violates modularity (routing is app business, not component)
3. The hassle to map the route to `initial states` is not worth it. Instead let the app map route to state on `didMount()` and apply it. No need for magic.

obuchtala:
@michael ^^

michael:
Totally agree. Will try to get this into the notes app today.

obuchtala:
You didn't look at the links, right? ;)

PR #568, opened 2016-05-01
Don't throw an exception when global.Substance is already defined, on…

jure:
…ly log a warning.

Given that this global variable is meant for debugging purposes only, perhaps it's enough to warn the user if it's already set, as a first step.

This warning will be seen by all users who have to separate Substance app instances on a page, so perhaps the second step should be to remove globals entirely.


michael:
That's fine like that. Thank you for the patch.

PR #571, opened 2016-05-04
Moving click target, the button target is way too small.

jure:
image

Only if you press on the very small ~25px x ~25px target will you fire the onClick event, clicking everywhere else will just close the popup without firing the event. This is completely unexpected. Clicking anywhere in the popup should fire the event.

### DIFF
--- a/ui/Tool.js
+++ b/ui/Tool.js
@@ -43,7 +43,9 @@ Tool.Prototype = function() {
   */
   this.render = function($$) {
     var el = $$('div')
-      .addClass('se-tool');
+      .addClass('se-tool')
+      .on('click', this.onClick);
+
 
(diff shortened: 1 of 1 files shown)


jure:
What's the status on this one, @michael? Any reservations?

michael:
Hm... I would rather solve this by making the button element expand fully, than changing the click target. Generally regarding Lens polishing we should dedicate a few days for it. Right after beta 4 is out.

michael:
However since I'm doing a tools refactor now, I will make this particular issue part of beta 4 :)

jure:
Cool, thanks!

michael:
Closing this. The toolbar stuff is now looking a bit different, so this fix no longer applies.

PR #577, opened 2016-05-15
Fix #576

### LINKED ISSUE #576: Clone props in $$ to avoid accidental changes to original data
Will follow up with a failing test.

michael:
See #576 - Combining props and children via append leads to a children property being added to props.

I observed this when working on the dynamic toolbar when passing props via reference and adding a child via append at the same time.

E.g.

``` js
$$(StrongTool, toolState.strong).append($$(Icon, {icon: 'fa-bold'}))
```

leads to `toolState.strong.children` being set. In debug mode I even get two VirtualComponent entries here as the manipulated props object is used for the second render.

@oliver---- Can you have a look?


obuchtala:
Wrong base.

PR #581, opened 2016-05-17
Isolated nodes

obuchtala:


obuchtala:
Even if this breaks the test-suite right now, we merge it to devel, so that we can continue work on tools and isolated nodes on shared source.

PR #582, opened 2016-05-19
Tools refactor

michael:


ui/Surface.js
```diff
@@ -191,6 +181,15 @@ Surface.Prototype = function() {
         this.clipboard.attach(el);
       }
     }
+    
+    if (this.props.overlay) {
+      el.append(
```
obuchtala: This is causing trouble, because ContainerEditor relies on the order of child elements when doing incremental updates.
Having this here, introduces an offset +1. So atm the wrong components get disposed, or components get inserted at the wrong position.

obuchtala: Is the overlay thing already part of this work actually?
I would say you remove that here and get this PR merged earlier.

michael:
Thx, will remove the overlay from Surface for now and merge.

PR #589, opened 2016-05-20
Document engine fixes

Integral:
Can’t really test it while so many things are broken with linter. But it should work this way as I have same tests against this codebase with mocha now in side project.


Integral:
Wrong branch.

PR #596, opened 2016-05-23
Implement a more robust version of getClientBoundingRect

### LINKED ISSUE #584: Implement getBoundingClientRect(els, relativeEl)
Working on overlay positioning I realise offset calculations etc. have not yet been implemented in a generic robust way. I propose the introduction of a utility method `getBoundingClientRect` that works very similar to native `Element.getBoundingClientRect` but:
- allows multiple elements being passed (we need this for selections that consist of multiple selection fragments)
- takes a `relativeElement` (a parent of els) that is used as a reference point, so we get coordinates relative to that parent, instead of relative to browser's viewport.

The following sketch illustrates a situation where we have a scrollpane (a container with overflow: auto) that consists of a `.content` element. Given an arbitrary scroll position and a selection (expressed as one or more spans) and we want to determine the bounding rectangle relative to the `.content` element. The blue lines shows the result of the native implementation, the yellow lines are the expected result.

image

Example usage within Surface:

``` js
var panelContentEl = this.context.scrollPane.refs.content.el;
var selFragments = nativeEl.querySelectorAll('.se-selection-fragment');
var rect = getBoundingClientRect(selFragments, panelContentEl);
```

Result should look like that:

``` js
{
  top: 700,
  left: 50,
  bottom: 500,
  right: 30,
  width: 320,
  height: 150
}
```

philippamarkovics:
This attempts to fix #584. Here’s what we need for this:
- [x] Split spanning elements into multiple client rects and calculate the union bounding box from them. This is needed if, for example, a `<span>` starts in the middle of a sentence but stretches over multiple lines. In that case, `offsetLeft` will report its offset only from where the `<span>` starts.
- [x] Find the union bounding box for multiple elements relative to a common parent and calculate its left/top/bottom/right offsets.
- [ ] Add tests for various scenarios like nested scrolling containers.

**Note:** This PR also adds a `toArray` utility function that’s used for converting `ClientRectList`s into proper arrays but can be reused for any non-Array list, e.g. `NodeList`.

**Naming**: I gave the function name some more thought and I think I would be for naming it `getRelativeBoundingRect`. I haven’t renamed it yet but I’d argue that the name should reflect that the bounding box (as well as the the offsets) is _relative_ to a parent element.

### DIFF
--- a/ui/Overlay.js
+++ b/ui/Overlay.js
@@ -57,20 +57,20 @@ Overlay.Prototype = function() {
 
     if (hints) {
       var contentWidth = this.el.htmlProp('offsetWidth');
-
-      // var contentHeight = overlayContent.htmlProp('clientHeight');
-      var selectionWidth = hints.rectangle.width;
+      var contentHeight = this.el.htmlProp('offsetHeight');
+      var selectionMaxWidth = hints.rectangle.width;
       var selectionHeight = hints.rectangle.height;
-      var containerWidth = hints.rectangle.left + hints.rectangle.width + hints.rectangle.right;
-      
-      this.el.css('top', hints.rectangle.top + selectionHeight);
-      var leftPos = hints.rectangle.left - contentWidth/2 + selectionWidth/2;
+
+      // By default, Overlays are aligned center/top to the selection
+      this.el.css('top', hints.rectangle.top - contentHeight);
+      var leftPos = hints.rectangle.left + selectionMaxWidth/2 - contentWidth/2;
 
       // Must not exceed left bound
       leftPos = Math.max(leftPos, 0);
 
       // Must not exceed right bound
-      leftPos = Math.min(leftPos, containerWidth - contentWidth);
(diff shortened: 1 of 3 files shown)


michael:
Looking good!

Few thoughts:
- span in the middle: maybe you can use span.getBoundingClientRect() there too, instead of el.offsetLeft? And then subtract the parent rectangles from it until relativeEl is reached? (but likely overlooking something here)
- toArray seems useful. just wonder it's necessary here as lodash/forEach should be able to iterate over NodeList ClientRectList etc. Haven't tested but is lodash/toArray able to do the same trick?
- top-center positioning of overlay: Fine with that. My only concern would be the edge case when you are selecting on the very top of a container and the overlay would not fit above the selection. I guess we would need to detect that case and position the overlay below.

👍  for getRelativeBoundingRect!

Let me know when this is ready to merge (we can push the test suite into a separate task) want to do naive click testing first. ;)

PR #599, opened 2016-05-25
Fix snapshot creation context.

Integral:
You missed context, this leading to server crushing.


Integral:
Wait, there is no `createSnapshot` method at all :) will fix

Integral:
Allright, ready to merge ;)

PR #602, opened 2016-05-29
Isolated nodes

obuchtala:
Next iteration.
- comply to new configurator API
- more idiomatic implementation: IsolatedNode and contenteditable states derived from single source


obuchtala:
This is how IsolatedNodes look like:

isolated-nodes-preview

obuchtala:
IsolatedNodes are easier to select and move around:
isolated-nodes-preview-2

PR #616, opened 2016-06-03
fix: table converters use setTagName() and setTagName() also copies textContent 

dmukhg:
- setTagName wasn't copying textContent.
- TableCellHTMLConverter and TableSectionHTMLConverter were using the older withTagName() method. This is no longer available in BrowserDOMElement

This was pull request #566, but had to reorganize my fork's branches.

### DIFF
--- a/ui/BrowserDOMElement.js
+++ b/ui/BrowserDOMElement.js
@@ -112,6 +112,12 @@ BrowserDOMElement.Prototype = function() {
     this.eventListeners.forEach(function(listener) {
       newEl.addEventListener(listener.eventName, listener.handler, listener.capture);
     });
+
+    var textContent = this.getTextContent();
+    if (textContent) {
+      newEl.setTextContent(textContent);
+    }
+    
     this._replaceNativeEl(newEl);
   };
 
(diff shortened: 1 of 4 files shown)


michael:
We have a new table implementation now in devel, so I think the table-related changes are obsolete. However can you rebase your other fixes to devel, I'm happy to merge them.

obuchtala:
@schatten While we are rewriting the table compent I will try  to pick your fixes in BrowserDOMElement

dmukhg:
Great!

Thanks for letting me know.

ui/BrowserDOMElement.js
```diff
@@ -115,7 +115,10 @@ BrowserDOMElement.Prototype = function() {
 
     var textContent = this.getTextContent();
     if (textContent) {
-      newEl.setTextContent(textContent);
+      // Warning, don't replace with setTextContent()
```
obuchtala, line 118: I hope not. Maybe there was another bug back then.
I.e., the element is created using `BrowserDOMElement.createElement()` which should indeed be a wrapped element.

obuchtala:
Closing this in favor of #744

ui/BrowserDOMElement.js
obuchtala, line 118: You were right, the bug was still there...Thx

PR #619, opened 2016-06-03
Switching from jshint to eslint

obuchtala:
eslint provides way more detailed control over which rules and how they are checked.
Also it has really valueable checks that reveal problems, such as better detection of missing `this` in closure etc.
On the other hand, it allows us safely to relax stuff like semi-colons, as eslint does check for the problematic cases, such as:

```
foo
[1,2,3].forEach(...)
```

Integration with Sublime is similar:
- install eslint with `npm install -g eslint`
- and use Sublime package manager to install the eslint package. You might need to go to the package settings and set the variables correctly, e.g. for me under Windows using `nvm` this was

```
{
  "node_modules_path": "%APPDATA%/nvm/v4.2.5/node_modules"
}
```

Edit: For me I have also changed the default keyboard short cut to what it was before in jshint


obuchtala:
@michael are u fine with that?
Next I will try to get the test-suite green again.

michael:
👍  Sounds good to me. I'll stay in Scientist-land... still reorganising and utlizing the config pattern where possible.

See: https://github.com/substance/scientist/issues/4

obuchtala:
@michael I am bringing back `console.warn` and `console.error` as with the redirecing util methods we lost the ability to jump to the source.
eslint will still warn us if we have `console.log` in the code.

obuchtala:
@michael almost there. Now only the collab tests are broken. Seems that the tests are not compliant with the protocol.

Edit: only `ChangeStore.test` and `CollabEngine.test`.

Integral:
Just FYI, on OS X sublime package settings:
"node_path": "$HOME/.nvm/versions/node/v4.4.4/bin",
"node_modules_path": "$HOME/.nvm/versions/node/v4.4.4/lib/node_modules"

PR #630, opened 2016-06-10
More fixes for IsolatedNodes.

obuchtala:


michael:
Nice!

I switched to a dedicated UnsupportedInlineNode implementation for scientist and made the XML editable. Just somehow managed to brake the inline-wrapper stuff, so figures are gona. Maybe you can have a look or we check it together.

michael:
Moving crossref editing into a modal as well, just less noise during the editing.

obuchtala:
Cool

obuchtala:
The UnsupportedIN work is on scientist?

michael:
yep. structured branch.

PR #639, opened 2016-06-12
Introduce configurable LabelProvider.

michael:


michael:
First iteration on switching from i18n to a configurable LabelProvider API. 

Currently usage via context (see SwitchTextTypeTool for an example)

``` js
// in render
var labelProvider = this.context.labelProvider;
var lbl = labelProvider.getLabel('delete');
```

Should consider making this a built-in for components so we can call `this.getLabel` in components.

Please review and merge.

obuchtala:
Looking good.. only linter thinks it's not ok ;)

obuchtala:
Will fix it.

Next time please run `gulp lint`.

obuchtala:
For classes please use the following pattern in future:

```
function MyClass() {
}
```

instead of 

```
var MyClass = function() {
}
```

The linter will honor this, by detecting invalid use of `this` pretty strictly. The only thing is, that we need to define constructors the way above.

PR #641, opened 2016-06-12
Added getLabel method to Component.

michael:
Ok to provide that short-hand method? We will use that a lot within render.


obuchtala:
Totally :)

PR #649, opened 2016-06-13
Fix documentation reader

michael:


michael:
Could need a bit of more love in terms of providing all config via configurator pattern, but will do for a beta 4 release I think. Please review and merge.

packages/prose-editor/ProseEditor.js
```diff
@@ -54,7 +54,7 @@ ProseEditor.Prototype = function() {
   };
 
   this._initialize = function(props) {
-    var configurator = this.props.configurator;
```
obuchtala, line 57: whoa.. des hat ned kracht die ganze zeit?

obuchtala, line 57: na... what is the difference to this.props?

obuchtala:
Fine with me to merge.. Though I probably need more exposure to grasp everything.
I.e. regarding generlizing TOCProviders.

packages/prose-editor/ProseEditor.js
michael, line 57: Ja es kracht bei willReceiveProps weil da nextProps reinkommen. Drum is ned aufgfallen.

michael:
Pushing the button...

PR #661, opened 2016-06-13
Get rid of i18n.

Integral:


doc/components/ExampleComponent.js
```diff
@@ -11,7 +11,7 @@ ExampleComponent.Prototype = function() {
   this.render = function($$) {
     var node = this.props.node;
     var el = $$('div').addClass('sc-example').append(
-      $$('div').addClass('se-heading').append(this.i18n.t('example'))
+      $$('div').addClass('se-heading').append(this.getLabel.t('example'))
```
michael: There's a typo here. Try to open the API docs and you will see.

doc/components/EventComponent.js
```diff
@@ -24,7 +24,7 @@ EventComponent.Prototype = function() {
           .on('click', this.onClick)
           .append($$('span').addClass('se-name').append(node.name)),
         $$('div').addClass('se-source').append(
-          $$('strong').append(this.i18n.t('event')),
+          $$('strong').append(this.getLabel('event')),
```
michael, line 27: You now need to provide the labels via addLabel. You can do this in doc/DocumentationReaderConfig.js just add some lines 

``` js
config.addLabel('event', 'Event'); // we don't need multi language here
```

michael, line 27: Actually see doc/i18n/en.js for all the translations. Just move them into DocumentationReaderPackage.js. And when done and things look good you can delete this doc/i18n/en.js

Same is true for substance/i18n/en.js + de.js. Just wipe it once all translations are done via config.addLabel.

michael:
👍

PR #680, opened 2016-06-15
Fixing InsertImageTool.

obuchtala:
Moved <input> dance from command to tool implementation


michael:
That looks better! Thanks.

PR #691, opened 2016-06-15
Allow provision of custom overlay class via config.

michael:


michael:
To change the overlay positioning for instance, subclass ui/Overlay and override `_position()`. Then in your EditorConfiguration do:

``` js
config.addComponent('overlay', MyOverlay);
```

PR #731, opened 2016-07-03
Coverage Reports

obuchtala:
- re-activated istanbul to create coverage reports
- after successful builds on travis they are pushed to coveralls: https://coveralls.io/github/substance/substance


Integral:
There is also coverify... Did you try it?

obuchtala:
Not worth to look at. Why the hassle. Istanbul is way more mature.

obuchtala:
BTW, `coverify` is basically a browserify transform plus a cli to create a report.
Istanbul plays well together with karma -- out-of-the-box. Being well documented and actively developed.

Integral:
Kk

PR #738, opened 2016-07-04
Drag and drop

obuchtala:
The idea is to have a global DragManager which will be the peer to register handlers, and which reacts on certain events.

There will probably be 4 major cases:
- dropping a file from outside
- dragging an isolated node
- dragging an inline node
- dragging via tool in overlay tools

Note: dragging selections directly is IMO not a good UX. I think if desired, a dedicated overlay tool should be used instead.

### DIFF
--- a/model/Document.js
+++ b/model/Document.js
@@ -523,6 +523,22 @@ Document.Prototype = function() {
     return new DocumentClass(this.schema);
   };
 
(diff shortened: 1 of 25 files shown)


obuchtala:
substance-example-image-drag-and-drop

obuchtala:
@michael could u have a look plz?

PR #743, opened 2016-07-08
Streamline configurator API's.

michael:
Move ProseEditor-specific API into separate ProseEditorConfigurator. Do not import firstPackage on construction. This is too early as inherited configurators may to things on construction before the first package is imported.

### DIFF
--- a/packages/prose-editor/FileClientStub.js
+++ b/packages/prose-editor/FileClientStub.js
@@ -1,10 +1,9 @@
 'use strict';
 
-var oo = require('../util/oo');
-var EventEmitter = require('../util/EventEmitter');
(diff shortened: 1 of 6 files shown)


michael:
Please review. I removed firstPackage argument from the constructor, so this change will break the examples. If you are good with this I will update the examples and then merge.

obuchtala:
👍

Integral:
Could you please explain new concept?

michael:
@Integral There is no new concept. It's just a realization that we should use multiple configurators to configure Reader/Writer separately. I need to get the examples updated and the PR's merged. I will be in the chat once done, so we can discuss the Notes app short.

PR #748, opened 2016-07-18
Fix deleting isolated nodes via `deleteCharacter`

philippamarkovics:
Currently, when deleting an isolated node via `deleteCharacter` (select node, press right arrow key, press backspace), various components (e.g. `SwitchTextTypeCommand`) fail because they’re trying to reference the deleted node.

This commit fixes it by inserting a new default node when the selection was a "full selection" — similar to how `deleteSelection` works.

@michael @oliver---- Is this a valid approach or does it need to be fixed somewhere else? I’m not sure I fully understand the root problem.


michael:
We will take a look in the next days. Thanks for the proposed fix.

obuchtala:
This has been changed completely.

PR #756, opened 2016-07-26
Fix SwitchTextTypeTool

nokome:
There seems to have been a recent change in the test type definitions which breaks the `SwitchTextTypeTool`. This is a proposed fix, using a mix of `testType.data.type` and `textType.name` in this tool's `render()` method.


michael:
You are right. Thanks for the proposed fix. Will check.

PR #758, opened 2016-08-03
Notes fixes

Integral:
@michael, @oliver---- pls merge.
Most notable change here is that documents are created now with 0 version, so editor will send transaction immediately with seed changes. It's what we did with factory functions on server before. Now we don't need any factory functions on server. It's fully compatible with old documents and I think much better. For example we could use different seed files within one setup.
I dropped out multiple schemas facility which we used before. I think it's good choice for now at least while we have single-schema-oriented configurator which we use now in document engine etc.


michael:
Looking good 👍

PR #762, opened 2016-08-08
(edit): Add scss files and add anon hash in server.js

slattery:
The PR has a few fixes that allow `npm install` and `npm start`.  The versions of browserify and watchify where chaanged to match the versions required in karma-browserify.  scss paths were fixed, or files added with scss paths.  In server.js the arg to serverJS was changed to a hash with key 'sourcePath' to match the structure of the serverUtils.serveStyles call.


michael:
Thank you!

We are currently overhauling the CSS files in general. We will no longer use SASS, instead provide just plain CSS for easier integration, and less dependencies. Changes will ship with Beta 5.

PR #764, opened 2016-08-15
CheerioDOMElement fixes and others

nokome:
Cheerio has a custom `type` attribute which it sets to `tag` for most elements but to `script` for scripts. This fix provides for consistent behaviour with `BrowserDOMElement`. Note that in both Chrome and Firefox `document.createElement('script').nodeType === window.Node.ELEMENT_NODE`

### DIFF
--- a/collab/CollabClient.js
+++ b/collab/CollabClient.js
@@ -41,9 +41,9 @@ class CollabClient extends EventEmitter {
   */
   _onMessage(msg) {
     if (msg.scope === this.scope) {
-      this.emit('message', msg)
-    } else {
-      console.info('Message ignored. Not sent in hub scope', msg)
+      this.emit('message', msg);
+    } else if (msg.scope !== '_internal') {
+      console.info('Message ignored. Not sent in hub scope', msg);
     }
   }
 
--- a/model/NodeSelection.js
+++ b/model/NodeSelection.js
@@ -112,6 +112,10 @@ class NodeSelection extends Selection {
     }
   }
 
+  _clone() {
+    return new NodeSelection(this.containerId, this.nodeId, this.mode, this.reverse, this.surfaceId);
+  }
+
 }
 
 NodeSelection.fromJSON = function(json) {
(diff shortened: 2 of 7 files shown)


nokome:
This PR now also has an additional, unrelated commit regarding the Websocket message scope console warning.

nokome:
Adds another commit around exporting images - only add the `data-preview-src` attribute if it's not null. This is perhaps a matter of taste - it's just the taste I prefer :wink: - don't add attributes unless necessary.

nokome:
Adds another Cheerio related fix

nokome:
@michael  : any chance some/all of these fixes can make it into beta 5? Let me know if you'd like me to rebase and/or create a new pull request/s.

michael:
Hey! Definitely. We just didn't get to bugfix mode yet. Rebasing them would be helpful yes.

nokome:
I rebased. Most of the commits above have some explanation as to the fix. You might not agree with them all.

michael:
Hey thanks. Maybe you could also squash the commits (and force-push to this branch), there are so many commits atm.

nokome:
Arrgh.  Yes, there are duplicates - I mucked up the whole rebase/merge stuff. Tried picking and squashing and that's wanting me to resolve the same merge conflicts again (like the 3rd time I've done it). So I might be better off doing `git reset --hard upstream/develop` and making those commits again manually.

nokome:
@michael : I sorted it out, thanks for your patience. For my future reference, I did:

```
git rebase -i upstream/develop
```

and `drop`ed unnecessary commits that were erroneously merged in previously and then did

```
git push --force
```

which I should have done the first time I rebased instead of `git sync`.

michael:
Thank you, we are having a look. I think the conditional import is not necessary. As for browserify, we look into that. The goal is that you should just be able to use a Substance release with browserify without the need to transpile.

michael:
Merged via #795.

michael:
Thank you!

PR #768, opened 2016-09-02
Back to pure CSS.

michael:
Removed SASS dependency.


michael:
Still missing bundling part for the docs but merging this into develop as a working version.

PR #769, opened 2016-09-07
Node selection clone

nokome:
This method was not implemented and was causing an error in Stencila Document's integration of Substance. All other classes derived from Selection have this method (e.g. ContainerSelection, PropertySelection) so it appears to be an omission.

Sorry, was attempting to cherry pick a single commit for this PR but failed, so it includes commits also in #764.


nokome:
Closing this as have cherry picked this commit into https://github.com/substance/substance/pull/764

PR #773, opened 2016-09-16
Remove ascii control codes from text/plain paste

tobiassodergren:


michael:
Merging this into draganddrop branch. Can you also rebase this to the develop branch, and submit another PR? Thank you!

tobiassodergren:
Hi Michael.

This fix does not work good enough to include in the develop branch, as is. The problem is that the DOM differs from the document model. To make this fix complete, the DOM needs to be updated to reflect what has been pasted into the document model.

Issue #798, opened 2016-09-29
API-Doc iteration


obuchtala:
From one scan over Texture we should revise the API Docs of these:
- [ ] AbstractEditor
- [x] Annotation
- [x] AnnotationCommand
- [ ] AnnotationComponent
- [ ] annotationHelpers
- [x] AnnotationTool
- [ ] Button
- [x] Component
- [x] Command
- [x] Configurator
- [ ] Container
- [ ] ContainerEditor
- [ ] DefaultDOMElement
- [ ] Document
- [x] DocumentNode
- [ ] DocumentSession
- [ ] DOMElement
- [ ] EditInlineNodeCommand
- [ ] EventEmitter
- [ ] FontAwesomeIcon
- [ ] Highlights
- [ ] InlineNode
- [ ] InsertInlineNodeCommand
- [ ] JSONConverter
- [x] Layout
- [ ] ProseEditor
- [ ] ProseEditorOverlayTools
- [x] ScrollPane
- [x] SplitPane
- [ ] TextBlock
- [ ] TextBlockComponent
- [ ] TextNode
- [ ] TextPropertyComponent
- [ ] TextPropertyEditor
- [ ] TOC
- [ ] TOCProvider
- [x] Tool
- [ ] Toolbar
- [ ] ToolDropdown
- [ ] \* all transformations

michael:
Take these first:
- Component
- Configurator
- DocumentSession
- Annotation
- DocumentNode
- DOMElement
- Document
- ProseEditor
- TextPropertyEditor
- ContainerEditor
- ScrollPane
- SplitPane

Then these:
- InlineNode
- TextBlock
- TextNode
- AnnotationComponent
- AnnotationTool
- AnnotationCommand
- EditInlineNodeCommand
- InsertInlineNodeCommand

dmukhg:
Component done and merged. #814 

Picking up Configurator now.

michael:
Hey! Good you are writing. Was about to take Configurator. ;)

I will work on documentation for all command/tool related stuff now (Command/AnnotationCommand, etc.)

Btw. @oliver---- is working on a much improved doc generator, so we should see some reward of our work soon. :)

dmukhg:
Great. I can take up Configurator, DocumentSession and Node related classes.

About the improved doc generator, that's good news!

michael:
@schatten the new doc generator is working now.. To serve and auto-compile the docs do:

```
npm run docs
```

[...]

PR #813, opened 2016-10-03
ui/Component: Document public API and overview #798

dmukhg:
This documents the main Component class and most of its public methods. I have used internal links liberally to ensure no matter where the reader starts reading, they can find their way around.


michael:
Merged. I just fine-tuned a bit.. See: #814. Thanks a lot!

PR #818, opened 2016-10-04
Extend EventEmitter on top level, so we don't need mixins.

michael:


michael:
Last commit breaks DOMElement.test.js.

```
not ok 746 .. and even the click handler should still work
  ---
    operator: equal
    expected: 1
    actual:   0
    at: Test.<anonymous> (/Users/michael/projects/substance/substance/test/ui/DOMElement.test.js:104:5)
```

@oliver---- How can I run the tests in the browser for debugging? Couldn't get this to work.

obuchtala:
Making DOMElement an EventEmitter is not correct, as DOMElement immitates the DOM event API.
1. We should find out why EventEmitter.mixin is not working as expected
2. Making ContainerAnnotationAnchor an EventEmitter is fine

obuchtala:
if you run `node make -s build test` the test-suite should be served

michael:
The problem is not the EventEmitter.mixin call. It's the EventEmitter.call(this) in the constructor.

> Class constructor EventEmitter cannot be invoked without 'new'.

obuchtala:
You don't need to call EventEmitter ctor anymore. It is a pure mixin now.

michael:
Ah ok.. then i just remove those calls. Thanks!

michael:
Ad running browser tests: I ran `node make -s build test` with a fresh install but get the error that test is not present. There's some invalid JS syntax there.

image

michael:
Ready to merge if green.

michael:
need to run for lunch.

obuchtala:
Rebasing this.

PR #826, opened 2016-10-04
Simplify Command API and added docs.

### LINKED ISSUE #825: Terminology used in Command implementations
I think these are a bit confusing at the moment.

Currently used terminology:
- params (passed in via ctor), they are specified at the configurator level. this is all static stuff
- props (passed in for getCommandState and execute) and contain documentSession, selectionState, surface, selection for getCommandState. For execute the previously computed commandstate is merged into props.
- context (provide access to app-specific childContexts, core tools don't use context)

Suggested terminology:
- config (relates it better to the configurator level and it's static nature)
- params (no confusion with props which are used in components)
- context (as is)

Instead of merging commandStates into params as done right now, I would provide commandState as an individual key. This would require some refactoring, but maybe worth to do that now as we can then freeze the command API for 1.0.

Thoughts?

michael:
Simplified command implementations to rely on params only. context is only needed if the user wants to access some app-specific objects and led to confusions, so I dropped it where it’s not needed.

@oliver---- pls review.

Fixes #825.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,7 @@ dist/doc
 tmp
 .npm
 .test
-.doc
+.docs
 
 # IDE / Editor genereated files
 .idea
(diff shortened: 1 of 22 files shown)


michael:
Also merging this, after some more tweaks.

PR #827, opened 2016-10-05
WIP: Docgen next

obuchtala:
To try out:

```
> git checkout docgen-next
> npm install
> node make docs -w
```

open `localhost:5550/doc/`

### DIFF
--- a/.docgenrc.js
+++ b/.docgenrc.js
@@ -1,4 +1,25 @@
 module.exports = {
-  "folders": ["model", "ui", "util"],
-  "repository": "https://github.com/substance/substance"
+  title: 'Substance',
+  // landing page
+  index: 'README.md',
+  // This will be used to create links to source files within the API docs
+  repository: "https://github.com/substance/substance",
+  sha: "develop",
+  content: [
+    { type: "page", id: "about", title: "About", src: "doc/about.md" },
+    { type: "section", id: "readme", title: "Readme", src: "README.md" },
+    { type: "section", id: "changelog", title: "Changelog", src: "CHANGELOG.md" },
+    { type: "page", id: "guides", title: "Guides", src: "doc/guides.md" },
+    { type: "section", id: "getting_started", title: "Getting Started", src: "doc/getting-started.md" },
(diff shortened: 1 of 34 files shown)


obuchtala:
Configuration is done via https://github.com/substance/substance/blob/89848c63458e6c27bc9483e13aa85eb7861948ed/.docgenrc.js

Right now everything is done on-the-fly, but we could certainly use the same code as a site generator.

michael:
Saugeil! Der build is verdammt schnell.

Makes it a lot more readable to have individual pages for classes, components etc. Only drawback that I now realize.. you can not search for any method name anymore.. so at some point I guess we need a search field and feed it with a static search index (just a json of name to url mappings should do). Or provide a everything on one page mode.

Issues:
- Deep linking to method names is currently not working
- E.g. in Component docs there is a link that resolves to: index.html#ui/Component#didUpdate

Questions:
- More a question: do we want to eliminate the folder prefix and and change `@link` statements to the flat address? I'd start doing that in new docs then.
- To do development, I have to npm link substance-docgen? Can you provide instructions?
- Does it make sense to use docgen for the whole substance.io site? E.g. the about page, could be what we currently have on the start page. That would require to implement a custom component, instead of providing a markdown as content source. Also no sections in the sidebar. This is more future, just thinking how it would affect the design of the generator.

To specify a custom page as component:

``` js
{ type: "page", id: "about", title: "About", src: $$(AboutPage) },
```

How could the design of the overall page be customized? E.g. maybe i don't want a top bar? Maybe a custom component could be provided here too.. and then there are only all needed props available... the pages information etc. (which page is active etc.) then you could render stuff however you want. Well, and we would have something like Jekyll but based on Substance Components ;)

Probably future.. this good what have so far. Will do css tweaks fill in more docs.

Is this mergable? (just needs small rebase because of daniels fix to index.es.js)

michael:
Will rebase and merge this now.

obuchtala:
Yeah, a lot to discuss.
I was also thinking to do the site-layout in a more Component.js way.
Let's see where we get, and would be good to collect some more ideas.

PR #833, opened 2016-10-05
Docgen fixes

### LINKED ISSUE #831: First line of class description is swallowed by doc generator
E.g. in Command.js

image

=> 

image

obuchtala:
- Fixed #831 
- Smart id lookup

I'd like to avoid using unqualified internal ids (such as `Component` instead of `ui/Component`) cause of name-clashes. We now have a LinkProvider which is mapping unqualified ids to qualified ones. As long there are no name clashes of the 'simple' names, this works as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ dist/doc
 tmp
 .npm
 .test
+.doc
 .docs
 
 # IDE / Editor genereated files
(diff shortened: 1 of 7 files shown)


obuchtala:
In case of name clashes we would need to fall back to qualified ids. In the core project we should not have this problem.

PR #835, opened 2016-10-06
Update paramaters when getting textType

khlbrg:
The `getTextTypes` method required access to the surface. So I just pass along the params object to the call


michael:
Good catch. Thanks for the fix!

PR #846, opened 2016-10-10
Documentation for record phase APIs of Configurator

dmukhg:
Because the Config interpreter APIs are not to be widely used by people using substance, not doing it right now.

### DIFF
--- a/util/Configurator.js
+++ b/util/Configurator.js
@@ -212,6 +212,14 @@ class Configurator {
     })
   }
 
+  /**
+    Adds an icon to the configuration which can be later retrieved via the 
+    iconProvider.
+
+    @param {string} iconName name or key for retrieving the icon
+    @param {Object} options  your custom method of representing the icon as a
+        JSON object. Enables plugging in your own IconProvider.
+   */
   addIcon(iconName, options) {
     var iconConfig = this.config.icons[iconName]
     if (!iconConfig) {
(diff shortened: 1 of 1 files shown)


michael:
Hi! Thanks for the additions. I read through the changes and they look very good!

Only one thing: The new doc generator is now able to resolve unqualified names in link statements. So instead of writing `{@link ui/Component}` we should use `{@link Component}`. Doing it that way allows us to move the code around, without breaking the links. For Substance we ensure that each class / module name is unique in the whole project. This also allows us to flatten exports and make it robust to structural changes in the codebase. Meaning that users can always

``` js
import {Component} from Substance
```

no matter where the actual file lives.

dmukhg:
Okay. Do you want me to update the links?

michael:
Please :)

dmukhg:
Updated the @links! 

BTW, do you have any tests for ensuring that classnames are not repeated? It is easy to miss and might break things without warning. Something like this 

`find collab doc model ui util | rev | cut -d '/' -f1 | rev | sort | uniq -d | wc -l` 

could work.

obuchtala:
@schatten I hope we won't need that in future, as I am aiming for a disambiguation mechanism, giving warnings when using an underqualified id, and preferring local matches. Though, this is not 'fully' implemented right now.

PR #863, opened 2016-11-14
Editing

obuchtala:


obuchtala:
Still buggy. But we should try to get it merged asap. Started a test-suite already. So it will be stable in short time.

michael:
Good stuff. Merge any time.

PR #871, opened 2016-11-15
Always fall back to text editing.

michael:
This is a quick fix for https://github.com/Infomaker/NPWriter/issues/11


obuchtala:
I think this is good enough for now. Thx.

PR #880, opened 2016-11-17
Fix for deleting non-mbp UTF-8 characters

tobiassodergren:
This fix handles the case where a character defined in a non-bmp area of UTF-8 is deleted (for example a smiley character). Characters defined on non-bmp is represented by 2 characters. Before this fix, the range to remove was incorrectly including one of the two characters.

### DIFF
--- a/ui/Editing.js
+++ b/ui/Editing.js
@@ -114,8 +114,8 @@ class Editing {
         nodeEditing.merge(tx, node, sel.start, container, direction, previous, next)
       } else {
         let start = offset
-        if (direction === 'left') start = start-1
-        let end = start+1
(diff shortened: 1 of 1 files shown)


obuchtala:
@tobiassodergren thx for this.

I suppose, we need to put that on our general agenda. I.e. this is probably not the only place where we need to be aware of it.

Thanks for pointing us to an entry point.

obuchtala:
Tracked by issue #885

tobiassodergren:
The proposed solution is not enough either, since characters may be a composition:

👨‍👨‍👧‍👧
Family (man, man, girl, girl)
Unicode: U+1F468 U+200D U+1F468 U+200D U+1F467 U+200D U+1F467

Issue #912, opened 2016-12-13
Surface.transaction is not a function


podviaznikov:
I'm seeing this error
```
SwitchTextTypeCommand.js:94 Uncaught TypeError: surface.transaction is not a function(…)
```
when trying to change text type on the text. I'm using latest version of the substance from the `develop` branch.

Not sure if you saw such an error previously. I mostly work from the `minimal` editor example but with the latest substance code.

michael:
I fixed that one https://github.com/substance/substance/commit/5ea129441939cf8ece454771819d50f8c7f4d484.

However, we are currently rewriting the editing behaviours, so don't consider develop stable atm. Better go with beta5 and then upgrade once we are ready.

podviaznikov:
Worked for me. Yes, I'm aware that develop is not stable. I'm just experiment and would like to try List feature

PR #914, opened 2016-12-14
Fix transaction is not a function

podviaznikov:
Fix similar to https://github.com/substance/substance/issues/912


michael:
Thank you!

PR #918, opened 2016-12-25
Use lodash-es.

michael:
Towards exposing Substance as a pure ES6 module. We should also provide substance-cheerio as an ES6 module so we can create single file bundles on the server as well.


obuchtala:
I just found out that lodash does not play well with tree-shaking.
https://github.com/rollup/rollup/issues/849
https://github.com/rollup/rollup/wiki/Troubleshooting#tree-shaking-doesnt-seem-to-be-working

The best way ATM is to use lodash-es but doing a 'deep' import, e.g.
```
import map from 'lodash-es/map'
```

rquast:
There's also this.. https://github.com/lodash/babel-plugin-lodash - I found the same problem with tree shaking and webpack (I'm using substance with webpack 2 instead).

PR #931, opened 2017-01-16
Fix getSelection is not a function

podviaznikov:
I often see following error
```
this.getSelection is not a function
    at ContainerEditor.onMouseDown
```

on this line https://github.com/podviaznikov/substance/blob/e28b9d6da97a83ffc844248072d34c4d3ade1b1e/packages/surface/Surface.js#L351

Adding binding to correct this.


michael:
Closing - we're addressing this separately.

PR #940, opened 2017-02-01
expose more utils

podviaznikov:
I'm using some of the utils in y custom `HTMLExporter`.
Substance already exposes a lot of utils. Not sure why those were left out.


michael:
Thank you!

PR #943, opened 2017-02-02
New tools/commands for lists

podviaznikov:
Up for initial review: Command and tool to create two types of lists.


michael:
Thank you! :)

You could pass an ordered property like so:

```js
config.addCommand('insert-unordered-list', InsertListCommand, { nodeType: 'list', ordered: false })
```

In the command implementation you can access it via `this.config.ordered`. Also see `ui/Command.js` for API.

podviaznikov:
@michael perfect! That what I was missing. Will update PR and have one command

podviaznikov:
Updated package. Not 100% sure with German labels. I've used Google Translate but not very confident with the results.

Also not sure how to test it? How do you test packages? Should I add PR into examples repo?

michael:
'unordered list' is 'Aufzählung', 'ordered list' is 'Nummerierte Liste'. Could you rebase your branch to develop? Then we have a cleaner merge. Thank you!

podviaznikov:
Updated, added few fixes too

michael:
Ready to merge. Only two requests:

- Rewrite commit history so no merge commits are visible (git rebase -i)
- Commit messages should start with upper case letter and end with ".". E.g. "Add label translations."

Thank you!

podviaznikov:
@michael done. Let me know if there are some other problems

michael:
Looking all good! Thank you!

PR #944, opened 2017-02-03
IsolatedNode refactorings

obuchtala:
- moved towards a simpler IsolatedNode structure: no 'brackets'/'slug', less markup.
- accordingly I adapted the selection mapping stuff.
- DOMSelection is now owned by AbstractEditor (not each Surface): this way DOMSelection mapping is generalized, at a wider scope, allowing for detecting switches between Surface via TAB for instance
- Some more fixes


michael:
👍

PR #945, opened 2017-02-03
Drag & drop implementation overhaul.

michael:


obuchtala:
Cool :)

Just found a bug, which I will fix: we should not set selection during initDrag

PR #948, opened 2017-02-05
Total Isolated nodes overhaul

obuchtala:
Still totally experimental


obuchtala:
@michael I found out, that having `draggable=true` on the IsolatedNode is impractical, as this produces problems in other browsers. My current approach is to use the blocker for that -- which is rendered only when the node is not selected :/
We need to discuss alternatives. E.g. rendering blocker also when selected (which leads to 3 clicks to get into an inner field), or maybe adding a draggable overlay (inside the isolated node) with drag handles.

michael:
The blocker solution sounds reasonable to me. While in some cases (images) - it may be desirable to drag an active node, in others it's better dragging is disabled (table had problems here, e.g. some areas between the cells were draggable)

Wondering if we should consider a double click to activate an isolated node. Then one could click (=select) and then drag, which is a common gesture.

Another possible way could be displaying some chrome (like as in Pages) once the element is activated. And have a drag handle (e.g. upper left corner).

obuchtala:
Right. These are the two best alternatives in my eyes.

obuchtala:
@michael Just wondering if we can ditch the extra support for nested isolated nodes. We have introduced it for the nested containers, to allow a 'seemless' UI, where u don't need to click through the levels. I don't believe that this is really useful. With Texture, we want to go for a 'flatter' editor ( ~author)

> E.g. this introduces the state 'co-focused'

michael:
I'd say ditch it. Unless we want to make building hierarchical XML editors a priority (cf. Texture publisher mode).

obuchtala:
Getting there :)

I managed to get two modes working: blocking (like the original isolated node) and open which does not have a blocker and uses 'contenteditable' isles which can be navigated seemlessly (like in beta-3).
Works well in Chrome and Safari.
In FF it is basically working too, but the 'open' mode is not supported as FF can't deal with contenteditable isles, and there are occasions where FF does not get 'out' using UP/DOWN, while LEFT/RIGHT works perfectly.

michael:
Wow super smooth! 

After some testing the only thing i'm not 100% sure about is the double click for activating isolated nodes (I guess users would expect a second click on a table cell should activate it). Maybe the explicit drag handle is better than the double click. We could still allow drags of non-active isolated nodes and only show the drag handle when it's activated.

obuchtala:
Yeah. True. I am shifting this though.

PR #949, opened 2017-02-05
Dragmanager fix

obuchtala:
Removes setSelection() from initDrag (which causes problems).

### DIFF
--- a/ui/DragManager.js
+++ b/ui/DragManager.js
@@ -63,7 +63,7 @@ class DragManager extends EventEmitter {
     ATTENTION: This can not be debugged properly in Chrome
   */
   _initDrag(e, options) {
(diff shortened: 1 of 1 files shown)


obuchtala:
@michael Just wondering how we could increase robustness of adding the dropzone layer. Are we sure that the dragzones are removed when something is going wrong? Does it work when user drags outside the editor?

obuchtala:
Just checked. Seems fine to drag outside the editor.

obuchtala:
Also checked: exceptions are no problems.

michael:
Yeah I also found it surprisingly stable ;) Especially now we can drag between multiple scroll panes and surfaces.

obuchtala:
Sweet :)

PR #960, opened 2017-02-10
Quote marks

obuchtala:


podviaznikov:
What does exactly this package do? Curious if I should use it

obuchtala:
Instead of inserting so called inch marks, it inserts left or right double quote marks.

PR #980, opened 2017-02-17
Surface fixes

michael:


obuchtala:
Hmmm. Obviously this is breaking something :/

michael:
I think it's the modifications of Component.mount. Looking into it.

michael:
No it's one of the InlineNode tests that breaks. We call editorSession.setSelection there, which somehow causes the surface to be disposed (and unregistered from SurfaceManager) where I made this change to null the selection. See: https://github.com/substance/substance/pull/980/commits/cea3a96d26642213c87abf298127cc88a8eff4ec#diff-499a04c6abee836cae4af3c9404b068eR67

Question: Should editorSession.setSelection ever cause a surface to be unregistered?


Here's the stacktrace:

image

PR #981, opened 2017-02-20
WIP: Container annotations revisited

obuchtala:


Integral:
Hey @oliver---- !
Can I do already something?
I saw working example. Meanwhile I could check performance on a big docs.

obuchtala:
Unfortunately I was not able to finish it yet.
What is missing is a good way to manage 'markers' (which are like the fragments before), for the nodes spanned by a container annotation.
So the 'anchors' get rendered, and transformed, but if you insert a paragraph it will not look like annotated until you reload.
You could indeed try out. It should at least improve your situation a bit.

obuchtala:
@Integral I have an idea about the markers
When I remember correctly, in Archivist we had brackets, and only showed the container annotation fragments when the bracket was selected. Right?
We could follow a similar approach, and generate the markers only for 'activated' annotations. 
Then it would be fine to re-create the markers when the container has changed -- which is not a good idea to do in general, as it is not efficient enough to do for all container annos.

Integral:
Yes, we did it like this. Only in case of editing or active state (but this
active state also belongs to reference, so when you selected topic X all
container annotations with reference to X got active). And the reason was
that otherwise we would have a big mess of colours instead of documents.
So this perfectly fine for what i am doing, but not sure that this would be
perfect for general usecase.

obuchtala:
Obsolete.

PR #987, opened 2017-02-22
Spreading arguments (surface) to super class

khlbrg:
AbstractScrollPane._onDomSelectionRendered is expecting a surface as a argument. Pass it from Scrollpane


obuchtala:
This has been fixed via #988

PR #990, opened 2017-02-23
Beta 6 fixes

michael:
Worked some more on overlay positioning. I now assign an id for the editorsession aka editorId. This I can later use to hide inactive overlays.

@oliver---- can you review and if good turn this into a new patch release? Then I could base the forms release on a substance release version.

### DIFF
--- a/model/EditorSession.js
+++ b/model/EditorSession.js
@@ -12,10 +12,12 @@ import MacroManager from '../ui/MacroManager'
 import KeyboardManager from '../ui/KeyboardManager'
 import forEach from '../util/forEach'
 import isPlainObject from '../util/isPlainObject'
+import uuid from '../util/uuid'
 import isFunction from '../util/isFunction'
 import isString from '../util/isString'
 import EventEmitter from '../util/EventEmitter'
 import FileManager from './FileManager'
+
 
 class EditorSession extends EventEmitter {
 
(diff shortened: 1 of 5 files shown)


obuchtala:
I am not too happy about the editorId caching. It does not look like the right approach. Maybe ScrollPane is not the right one to aggregate all the hints necessary for positioning.
So it seems, that for positioning there are different hints needed, one being the selection rect, one the editorId, and maybe more. Isn't the editorId and the selectionRect very much related?

obuchtala:
I am against rolling out this experiment. Until we have found a proper way we need to make forms work with a sha, or maybe via a `tar.gz`.

obuchtala:
We should put this into a branch with different name. It is not about fixing beta-6.
More like adding new features.

michael:
OK let's wait with this then and further discuss. There's another thing I find a bit weird: When I create the Overlay instances, I need to assign the editor as the parent in order to make the context available to the overlay component. Since the overlays are injected in the body scrollpane it's weird to have its parent set to the editor component. Possibly all needs more thinking. At least everything works smoothly now for forms (from a user point of view).

obuchtala:
@michael is there a way that you base substance-forms on beta-6.4 and add code to substance-forms which monkey-patches the core classes (i.e. overwriting the prototype impls with custom versions)?

michael:
I think it's fine to base it off a sha.

PR #991, opened 2017-02-23
WIP: Overlay positioning

michael:
Improved API's for overlay positioning.

### DIFF
--- a/model/EditorSession.js
+++ b/model/EditorSession.js
@@ -564,7 +564,9 @@ class EditorSession extends EventEmitter {
     @internal
   */
   startFlow() {
-    if (this._flowing) return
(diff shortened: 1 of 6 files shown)


michael:
Discuss how we can avoid calling this.context.scrollPane in Surface and editorId caching in Scrollpane is not nice.

obuchtala:
What is actually missing ATM, is a layer above AbstractEditor/EditorSession.
A cleaner approach would be to store positioning hints in a shared scope and use flow stages to execute at the right time. 
Right now EditorSession is 'owning' the flow, thus we can't solve the problems here yet.

obuchtala:
Obsolete.

PR #994, opened 2017-03-01
Beta 6 fixes

dannelundqvist:
When using US international keyboard on windows dead keys/combined characters are not handled correctly. An example is a single straight quote that is the combination of two key presses where the last key is space. The dead key can not reliably be identified by keycode 229 on windows using US intl keyboard.

For this to work Substance should look for event.key "Dead" and it should not handle the space key itself. This is also true for several international characters (ü) that is created using a combination of dead keys and another key.


obuchtala:
Great!
I will try this out a bit.

PR #1010, opened 2017-03-22
Improve rendering

obuchtala:


obuchtala:
@michael as u suggested I started adding to CHANGELOG.md
Let's make this best-practice, finishing a PR with a CHANGELOG update

PR #1020, opened 2017-03-28
Overhaul

obuchtala:
Aims:
- Grand cleanup (removing dead code, following best practices, etc.)
- Increasing test coverage (>80% on each 'module')

Why?
We want our test-suite be more representative. For that, we now use the bundled substance library from within tests, eating our own dog-food.
substance is now mature enough to flesh out the test-suite and reach a reasonably high test-coverage.

Status:
- [x] `/dom`
- [x] `/model`
- [ ] `/packages`
- [ ] `/ui`
- [ ] `/util`


obuchtala:
Let's call it a first round. `dom` and `model` looking good.

`ui` and `util` must be covered better in a second round.
`packages` and `collab` are utterly undertested.

PR #1019, opened 2017-03-28
Added config option to force plain-text paste

tobiassodergren:
Pasting HTML gives unwanted results. This pull request enables editor sessions to be configured
to allow plain-text paste only, except for in-document copy-paste operations.

PR #1021, opened 2017-03-30
Added config option to force plain-text paste

obuchtala:
Pasting HTML gives unwanted results. This pull request enables editor sessions to be configured
to allow plain-text paste only, except for in-document copy-paste operations.

(This was #1019 before)


obuchtala:
@tobiassodergren This makes definitely sense.

Issue #736, opened 2016-07-04
Add classical text macros


obuchtala:
Markdown inspired macros are pretty common and should be added to our packages:
- `#` (at begin of a paragraph): turn into a heading
- `*` or `-` (at begin of a paragraph): turn into a list item
- `**..**`: make strong
- `*..*`: make emphasis

Make packages configurable so that these can be skipped if not wanted.

PR #1023, opened 2017-04-07
added support for h2, h3 and paragraph macro

PookMook:
first jab at thinking at #736 (only for headings promoting + paragraph demoting, list and stuff comming up soon)
```
# => h1
## => h2
### => h3
#### => paragraph (is not a markdown spec, but there should be a way to demote to paragraph, isn't it?)
```

Do you prefer to put this in a different module, as this one might be used already in different project (and you don't want to change behavior of it) ?


obuchtala:
Obsolete.

PR #1026, opened 2017-04-11
Fixes

michael:


michael:
@oliver---- this fixes some things to make Forms work with the latest Substance. However, annotations (commands) are broken (tools don't show up)

I created a branch `latest-substance` for forms that you could check out and check if you have the time.

PR #1038, opened 2017-05-17
Collab rebase fix

tobiassodergren:
Fix for rebase functionality, which seem to read from a property that was always undefined. Also, the version that was returned from the call to getChanges was missing in the callback


michael:
Thank you!

PR #1040, opened 2017-05-19
WIP: Find&replace

michael:


test/FindAndReplaceManager.test.js
```diff
@@ -27,10 +27,10 @@ test("Find matches and select first after current selection", function(t) {
   editorSession.setSelection({ type: 'property', path: ['p1', 'content'], startOffset: 7, endOffset: 7 })
   manager.startFind('123')
   t.equal(manager._state.matches.length, 4, 'Should have 4 matches selected')
-  t.equal(manager._state.selectedMatch, 1, 'Second match should be selected')
+  t.equal(manager._state.selectedMatch, 0, 'Second match should be selected')
```
michael: With this test I wanted to check if the next match is found according to latest cursor position before the search started. So if you have your cursor at p1:7 and start a search p2:1..4 should be selected.

michael: This fix is fine actually. Follows the same semantics as SublimeText.

michael:
I'm closing this opening a new PR for new branch (find-and-replace)

PR #1046, opened 2017-05-31
Find and replace

michael:


packages/find-and-replace/FindAndReplaceManager.js
```diff
@@ -209,20 +219,24 @@ class FindAndReplaceManager {
   }
 
   _propagateUpdate(renderSelection) {
+    // HACK: we make commandStates dirty in order to trigger re-evaluation
+    this._updateMarkers()
+    this.editorSession._setDirty('commandStates')
+    if (!renderSelection) {
+      this.editorSession.skipNextSelectionRerender()
```
obuchtala: I would prefer to implement this in the way I explained yesterday:
- introduce a 'focused' or 'blurred' state
- skip selection re-rendering if blurred

michael: Ok. I was just concerned that you may end up trapped if some implementer forgets to unblur right after the performed action.

obuchtala: Just don't think skipping next is enough. E.g. think if a reflow happens because of collaboration stuff. Then you would loose the cursor too.
I think the only valid implementation is to have an explicit app state. Even if that means that the app state needs to be managed carefully.

Integral:
Yay! Let's merge at last!

michael:
as long as it is green ;)

Integral:
Hurrah!

PR #1079, opened 2017-07-14
Command states active with Surface.props.command

nmcfarl:
Change the include.map to return the name, not true or false, thus enabling toggle buttons on surfaces with surface.props.commands


obuchtala:
Closing this in favor of #1081

PR #1081, opened 2017-07-17
Fix command state updating.

Integral:
With #1078 you’ve introduced serious regression, a lot of commands not working at all (at least select all, undo, redo), but I believe that all of them was broken.
I think that you wanted to do a filtering of white-lists commands, but instead did a wrong thing.
Would be good to create a test case to prevent this kind of mistakes in future.

### DIFF
--- a/ui/CommandManager.js
+++ b/ui/CommandManager.js
@@ -111,7 +111,7 @@ export default class CommandManager {
       let included = surface.props.commands
       let excluded = surface.props.excludedCommands
       if (included) {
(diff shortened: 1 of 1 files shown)


nmcfarl:
This is a better written version of my pull request at https://github.com/substance/substance/pull/1079   - I'd say you use this unless there's a good reason to use the map and not a filter.

obuchtala:
Sorry for the regression. Agree it should be tested.

PR #1082, opened 2017-07-17
Do not assign non-existing ids for annotations.

Integral:
Currently when you pasting annotated text (let's say HTML one) only first paragraph got annotated.
That's happening typically because you are assigning non existing ids to annotations (at least when you are pasting content inside empty document).


obuchtala:
Would be good to have a test case revealing the problem.

Integral:
When a lot of red is a good sign :)
I will look closely and try to make it green, #1083 is here and other things also I believe...

obuchtala:
Haha. It seems you really broke it ;)

obuchtala:
Let me know if I can help :)

obuchtala:
Oh, I see... you added a revealing test.
Very good!
Let's fix it. Let me know if I can help you.

Integral:
I will do it today ;)

ui/ClipboardImporter.js
```diff
@@ -90,8 +90,37 @@ class ClipboardImporter extends HTMLImporter {
     // specific format, e.g., id="docs-internal-guid-5bea85da-43dc-fb06-e327-00c1c6576cf7"
     let bold = body.find('b')
     if (bold && /^docs-internal/.exec(bold.id)) {
-      return bold
+      body = bold
     }
+
+    body.findAll('span').forEach(span => {
+      const style = span.el.style
```
obuchtala: In general you should use span.getNativeElement() instead of span.el as  does not work in the memory DOM.

obuchtala: ... but as you see below, accessing the native element here is not a good idea

ui/ClipboardImporter.js
```diff
@@ -90,8 +90,37 @@ class ClipboardImporter extends HTMLImporter {
...
+      // Google Docs uses spans with inline styles 
+      // insted of inline nodes
+      // We are scanning each span for certain inline styles:
+      // font-weight: 700 -> <b>
+      // font-style: italic -> <i>
+      // vertical-align: super -> <sup>
+      // vertical-align: sub -> <sub>
+      let nodeType
+      if(style['font-weight'] === '700') {
```
obuchtala: You should use getStyle('font-weight') here. This works in browser and nodejs.

ui/ClipboardImporter.js
```diff
@@ -90,8 +90,37 @@ class ClipboardImporter extends HTMLImporter {
...
+        nodeType = 'b'
+      } else if (style['font-style'] === 'italic') {
```
obuchtala: same same

ui/ClipboardImporter.js
```diff
@@ -90,8 +90,37 @@ class ClipboardImporter extends HTMLImporter {
...
+        nodeType = 'i'
+      } else if (style['vertical-align'] === 'super') {
```
obuchtala: span.getStyle('vertical-align')

ui/ClipboardImporter.js
```diff
@@ -90,8 +90,37 @@ class ClipboardImporter extends HTMLImporter {
...
+        nodeType = 'sup'
+      } else if (style['vertical-align'] === 'sub') {
```
obuchtala: dito

obuchtala:
I have the feeling that it is very hard to support these combined formatting annotations.
Maybe we accept for now, that there will be more fragments than necessary, and relax the test-case a bit.
On the long run we should try to improve the result by fusing adjacent annotations of the same type.

Integral:
@oliver---- there are new problems. And the main one is that we have same annotation type overlapping. Is that ok generally that we have two strongs on some range?

obuchtala:
It is ok I think. In general we can not avoid this, so we need make sure to be able to deal with it.

obuchtala:
@michael is having two bolds at the same place a problem for the UI?
As I said, this can actually come in from XML/HTML, and is valid content there.

Integral:
God! Merge it pls while it green!
We definitely need to tidy html in future. That's the only independent from editor way...

obuchtala:
Hehe :)
I'll take another look first, and then ...

obuchtala:
👍

obuchtala:
@Integral Good work. Thx!

PR #1087, opened 2017-07-21
Use flexbox for styling ScrollPane.

michael:


michael:
This causes problems in Firefox, which we need to investigate first.

obuchtala:
Obsolete.

PR #1095, opened 2017-07-28
WIP: Container annotation manager.

Integral:
This is not yet ready to merge, but in the same state as it was in February with all new changes.
Please review.

### DIFF
--- a/packages/container-annotation/ContainerAnnotationManager.js
+++ b/packages/container-annotation/ContainerAnnotationManager.js
@@ -22,7 +22,12 @@ class ContainerAnnotationManager {
 
     this._containerFragments = {}
 
-    //this.initialize()
+    // TODO: avoid this ugly hack
+    // We can't call it imediately as there is no Marker Manager yet
+    // We should somehow track when editor session did initialised
+    setTimeout(() => {
+      this.initialize()
+    }, 100)
   }
 
   dispose() {
@@ -33,8 +38,8 @@ class ContainerAnnotationManager {
     this._computeAnnotations()
     this._updateAnnotations()
     // HACK: we make commandStates dirty in order to trigger re-evaluation
-    // this.editorSession._setDirty('commandStates')
-    // this.editorSession.startFlow()
+    this.editorSession._setDirty('commandStates')
+    this.editorSession.startFlow()
   }
 
   getAnnotationFragments(annoId) {
--- a/packages/container-annotation/ContainerAnnotationManager.js
+++ b/packages/container-annotation/ContainerAnnotationManager.js
@@ -1,26 +1,48 @@
 import { Marker } from '../../model'
 
-class ContainerAnnotationManager {
+/*
+  This class maintains a set of Markers (see MarkersManager)
+  where each marker represents a fragment of a ContainerAnnotation.
+
+  It needs to react on DocumentChanges whenever:
+  - a ContainerAnnotation has been created, deleted,
+    or updated
+  - a node has been inserted or removed from a container
+    being covered by ContainerAnnotation
+
+  TODO: find out if this requirement description is complete
+
+  > We should discuss, if it might be helpful to restrict ourselves
+    to a set of manipulations (document changes) with a specific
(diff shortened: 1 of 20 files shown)


packages/container-annotation/ContainerAnnotationManager.js
```diff
@@ -0,0 +1,164 @@
...
+          // conntainer annotation got updated
+          if(this._state.annotations[nodeId]) shouldUpdate = true
+        }
+      }
+    })
+
+    // Compute range for created/removed annos
+    if(dirtyNodes.length === 2) {
+      // TODO: We should get container via API
+      const container = this.doc.get('body')
```
obuchtala, line 104: We need to fix this.
I need to take a closer look at this method. Maybe I have an idea.

obuchtala: Will let u know if I find something.

Integral: Also maybe we should move node range computation to container annotation model.

michael:
@Integral please in future send separate PR for things like SnapshotEngine fix, then I'm able to get them into develop faster.

Integral:
@michael i don't know what it's doing here... i should remove this, because it was fixed by me right in develop branch.

michael:
@oliver---- we are publishing this as an npm release `1.0.0-container-annos.1` and `npm publish --tag container-annos`

However it would be good to get this into develop. Do you think it would affect stability.. i mean if no container annos are present in the document.

obuchtala:
I guess we want this merged sooner than later, to be able to find out :)

michael:
So should we do it? :)

obuchtala:
I guess so :)

michael:
Anything to clean up or review before we do it?

obuchtala:
There are still some brutal hackz, like `doc.get('body')` hard-coded.
We need to fix that first :(

Integral:
@oliver---- you was going to think about it ;)

michael:
Hmm... how to best get rid of that hardcoded body?

obuchtala:
Need to understand the implementation. But this is not what will happen instantly.

obuchtala:
Let us put it on the agenda.

obuchtala:
We need to rewrite `ContainerAnnotationManager._onDocumentChange()`.
I have started to do some tidy-up in that class, to increase comprehensibility.
Now we have to identify a viable strategy how to detect changes that affect the indexes maintained by
this class.
And we need a test-case for this.

obuchtala:
There are tests in `ContainerAnnotation.test.js`

obuchtala:
I left some comments:
>   This class maintains a set of Markers (see MarkersManager)
> where each marker represents a fragment of a ContainerAnnotation.
>
> It needs to react on DocumentChanges whenever:
> - a ContainerAnnotation has been created, deleted,
>   or updated
> - a node has been inserted or removed from a container
>   being covered by ContainerAnnotation

>  TODO:
>  - try to complement the requirement description started in the class header
>  - try find a reasonable implementation of the description
>  - maintain an explanation of the taken approach here, so that
>    it is easier to understand and verify the implementation in future

Integral:
@oliver---- we don't want to merge until those todos will be done?

obuchtala:
Nope. Sorry.

obuchtala:
Because the applied hackz are so brutal, that we must solve the proposed TODOs anyway

Integral:
Fine for me as we did npm release anyways and i will not be able to dive into details until Friday.

obuchtala:
Ok

obuchtala:
Some ideas:
- Detecting changes to ContainerAnnotations should be rather easy: there are several ways to approach this, either via inspecting `change.ops`, or by checking if `change.updated` or `change.deleted` affect known annos, plus looking into `change.created` if there is container anno. Probably the first is easier.
- Detecting changes to Containers is probably more involved: on the highest level (without lists) it is rather easy by detecting updates to containers. For lists, we would need to identify updates to lists and navigate up to the container.

obuchtala:
Obsolete.

PR #1104, opened 2017-08-08
Collab fixes.

Integral:
Here will be fixes for current collab protocol.


Integral:
@oliver---- pls review one line

obuchtala:
Will do

obuchtala:
Looks like it was a regression. 👍

Issue #1111, opened 2017-08-14
Pasting few paragraphs in multiline editor


Integral:
We are not covering a case now when someone paste few paragraphs into multiline editor ([e.g. pasteDoc, but without Container](https://github.com/substance/substance/blob/develop/model/paste.js#L29)).
This turning into error later when we are trying to [get position of node inside non-existing container](https://github.com/substance/substance/blob/develop/model/paste.js#L120).

My proposal is to have a few options:
1) Have an option for TextPropertyEditor to always paste plain text 
2) Merge nodes with new lines between them (could be quiet difficult actually)

@oliver---- what do you think?

Integral:
While we have no time to tackle this seriously we could merge #1119. It will paste plain text and not throw errors at least.

obuchtala:
I like your solution, because it improves the situation in any regard.
The other complex solution sounds a bit too much hassle for the rather low benefit.

obuchtala:
Fixe via #1119

PR #1119, opened 2017-08-23
Pasting few paragraphs in multiline editor.

### LINKED ISSUE #1162: Pasting HTML into TextPropertyEditor
If you paste multiline HTML into TextPropertyEditor it will produce error [here](https://github.com/substance/substance/blob/2e5320993f62e51b44f82bf97c2ce29e854b991d/model/paste.js#L125) as there would be no containerId. 
I guess we should check if selection have containerId during pasting and if not paste plain text version.

Integral:
As described in #1162 #1111. Naive implementation to insert plain text and not run into errors.

This fixes #1162


obuchtala:
@Integral Thanks. Will pick this up, rebase, and refine.

obuchtala:
@Integral take a look how I like such things to be fixed :)
I.e. adding a test revealing the problem, and then fixing it.

obuchtala:
Fixes #1111 too

PR #1120, opened 2017-08-29
Improvements on the XML API

obuchtala:


michael:
In the process of upgrading Texture to the improvements branch. Had to make a fix to ensure defaultTextType gets passed to the schema config if available. 

There are a number of regressions when it comes to editing:

- breaking a text node copies the whole text to the new node (instead of the trailing text after cursor)
- no re-render when annotations are created/removed
- TOC in Texture is no longer computed properly (needs investigation)

You can observe when running Texture against the improvements branch.

Question: How can I use the new hierarchical change listeners? They would be really useful in Texture.

solsson:
Is there a particular git branch/commit that is perferrable for integrating plain old Substance now?

What's the status of the APIs, compared to an anticipated 1.0.0? Not only XML, but also:
 * JSON serialized format
 * Transaction format
 * ImageFileProxy
 * Preconfigured setups like ProseEditor

obuchtala:
@solsson The most things will remain the same. The whole XML oriented implementation is just an opt-in and considered the post 1.0 track.
Re: ProseEditor - it will remain too, at least for 1.x. However, on the long term we are considering to pull this out and provide it as extra bundle, particularly as an HTML rich-text editor.

obuchtala:
Want to merge this to finish this iteration and provide a preview bundle.

solsson:
@oliver---- Great! Will you npm publish `1.0.0-preview.2`?

Issue #1121, opened 2017-09-02
Basic support of internet explorer


Integral:
By "basic support" I mean lack of errors during script loading.
Currently any file with substance.js will produce blank page in IE11. While es5 bundler is broken (you have to change condition [here](https://github.com/substance/substance/blob/develop/make.js#L96) to `target === 'browser:legacy'`), the real problem is that es5 bundle will have at least `Object.assign()` method which isn't supported by IE at all.

Integral:
Looks like buble is very much sensible... Currently it produced very strange code. Don't know what to do with it...

obuchtala:
Obsolete.

PR #1122, opened 2017-09-02
ES5 transpiler.

Integral:
Fix for ES5 transpiling. As I mentioned in #1121 we also need to include some polyfills. At least for `Object.assign()` and `Array.from()`. Maybe it’s better to do that via Substance Bundler, for instance with special option `polyfills`.
Still we need to take a closer look at results of Buble compiler. I have a feeling that something goes wrong there which could be true, as [it is not working with all features](https://buble.surge.sh/guide/#unsupported-features).

### DIFF
--- a/make.js
+++ b/make.js
@@ -95,7 +95,7 @@ function buildLib(target, production) {
       ]
     }
   }
-  if (target === 'legacy') {
+  if (target === 'browser:legacy') {
     config.buble = true
   }
   if (production) {


obuchtala:
You're right. bublè is not complete. I think we do not use features that are not transpiled by buble.

tamlyn:
Any reason this hasn't been merged?

obuchtala:
Nope. Will do it asap.

obuchtala:
Thanks @Integral!

obuchtala:
Will publish.

obuchtala:
Published as 'substance@1.0.0-preview.54'.
@tamlyn Please let me know if that fixed it

Issue #1128, opened 2017-10-10
Container placeholder


Integral:
Hey!
We've got now kind of strategy for placeholder rendering inside container editor, the logic described [here](https://github.com/substance/substance/blob/develop/ui/ContainerEditor.js#L272-L287).
But at the same time all text nodes are receiving the same placeholder property and renders it in case of empty text. 
E.g. if you passing placeholder it will be shown in every empty paragraph:

![screen shot 2017-10-10 at 11 55 09](https://user-images.githubusercontent.com/182010/31377800-ea0ad63c-adb1-11e7-8088-95a9318bbd17.png)

I see a one way to fix it. Let's maybe rename container placeholder to containerPlaceholder and not mix it with text nodes placeholders?

michael:
Hmm.. the idea here was to only render the given placeholder if there's exactly one empty node. If you have multiple nodes, no placeholders should be visible.

Integral:
Yes, but [here](https://github.com/substance/substance/blob/develop/ui/ContainerEditor.js#L120) you render it for each text block.

michael:
only `this.placeholderNode` should have placeholder property set. Must be some error somewhere if not the case.

Integral:
It's very easy. For each text node we are calling ```super.renderNode($$, node, nodeIndex)```, right? We are not passing placeholder here to the renderer.
But take a look at [renderNode method](https://github.com/substance/substance/blob/develop/ui/Surface.js#L126-L139). It is using the same ```this.props.placeholder```. Didn't you see? We are using the same placeholder which we passed to container editor in parent class. So it doesn't really matter what we are doing inside container editor.
We could also override surface `renderNode` method to take placeholder from node.

michael:
Hmm yeah i guess we should not pass through placeholder there, as it's set afterwards explicitly in _attachPlaceholder. Guess you could remove that?

Integral:
I don't want to remove it from Surface, it's dangerous, don't you think so? Let's override `renderNode` on container editor level, ok?

michael:
Ah you are right. We want to pass it through in the case of the TextPropertyEditor but not ContainerEditor. Pls fix as you suggested.

Integral:
@michael ready!

Integral:
Can I merge?

PR #1129, opened 2017-10-10
Render container editor placeholder only for one empty node.

Integral:
As discussed in #1128


michael:
Maybe there's a way to avoid code duplication. E.g. moving the minimal portion into a `_extractNodeProps()` function and then override that instead of full `renderNode()` function.

Integral:
@michael pls check again

PR #1134, opened 2017-11-19
Update node with new properties

michael:


michael:
@oliver---- is this a valid approach?

model/Node.js
```diff
@@ -212,7 +212,8 @@ function _compileSchema(schema) {
 function _compileDefintion(definition) {
   let result = definition
   if (isArray(definition.type) && definition.type[0] !== "array") {
-    definition.type = [ "array", definition.type[0] ]
+    definition.targetTypes = definition.type
```
obuchtala, line 215: Yep. I think this is perfect.

Issue #1092, opened 2017-07-28
Possible cyclic (circular) dependencies causing issues with Webpack


rquast:
I've been using webpack to bundle substance. It worked with beta 6 until I tried beta 7.

I managed to get it to work again by changing the following:

```
import { getSelectionRect, getRelativeMouseBounds, platform } from '../util'
import { DefaultDOMElement } from '../dom'
import Component from '../ui/Component'

class AbstractScrollPane extends Component {
```
I changed import { Component } form '../ui' to skip the re-export.

I've tried checking for cyclic dependencies with madge, but I think it crashes each time it hits one.

obuchtala:
Somehow `Rollup` is able to deal with the code. At the beginning we had to resolve real cycles such as  `Selection.js` having a factory method depending on `PropertySelection.js` which apparently extended `Selection` exported by `Selection.js`.
`webpack` seems to be choking with the `index.js` files.
Have you tried to separate webpack and Substance bundling? I.e. using the single file bundle?

rquast:
Yes, I tried to use the single file bundle, but for some reason or other, it's very slow for webpack to process. I might try forking the project and making a webpack build to see if I can recreate the problem, at least to test new builds.

Other than that small thing, everything works perfectly.

obuchtala:
Ok, it's slow, because it really needs to analyse the whole library, and can not drop things on the fly.

obuchtala:
Maybe it is an option for you to load `substance.js` via an extra `<script>` and tell webpack to treat it as external located at `window.substance`?

Edit: this is what we do in Texture, too

rquast:
That's probably a good idea too. I wasn't really keen on splitting any libraries from the bundle, but the editor is a good exception because of the size.

PR #1147, opened 2018-02-03
Fix imports to be more ganular, avoiding index.js imports

kilburn:
After trying to compile substance using webpack, I hit bug #1092. This PR updates all of the internal imports to be fine-grained, without depending on the various index.js re-exports.

This solves the cyclic issue during compilation and should help generating smaller customized versions of the editor.

All test are passing, but during compilation there are two warnings that didn't exist before:
```
> eslint -c .eslintrc.js collab dom model packages test ui util xml index.es.js

Rollup: /Users/kilburn/Documents/Projects/tu/substance/index.es.js -> /Users/kilburn/Documents/Projects/tu/substance/dist/substance.es.js
The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten
https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined
The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten
https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined
..finished in 7675 ms.
```

There is no reference to any specific file in the messages and I don't know rollup well enough to discover where are these "problematic" `this` usages, so I cannot pinpoint if this is an actual problem or not.


obuchtala:
@kilburn thank you for your PR
I need to check if everything is fine locally.

Similar as in #1092 I would still recommend to work with the published bundled version of substance. Latest versions of substance are published using tag 'preview'. We usually install it via explicit version, e.g. the current is `substance@1.0.0-preview.21`.
In the dist folder of that there are several bundled files already. You could use the browser bundle directly. The Nodejs one is looked up automatically via package.json. Additionally you find an es6 bundle, `dist/substance.es.js` which could be used further and reduced using 'tree-shaking'.

E.g. this webpack config worked for me:
```
const path = require('path');

module.exports = {
  entry: './index.js',
  output: {
    filename: 'bundle.js',
    path: path.resolve(__dirname, 'dist')
  },
  module: {
    loaders: [
      {
        test: /\.js$/,
        exclude: /node_modules/
      }
    ]
  },
  resolve: {
    alias: {
      'substance$': path.resolve(__dirname, 'node_modules/substance/dist/substance.es.js'),
    }
  }
};
```

obuchtala:
BTW, the explicit import is not the solution to more minimal bundles. This is related to 'tree-shaking'. The problem is that there are some patterns that existing tree-shaking implementations do not consider as side-effect free. For example, something like this:
```
class MyClass {}
MyClass.foo = 'bla'
```
This has not been considered side-effect free, even if it was done only in the same module. So `MyClass` appeared in the bundle no matter if you were using it or not.
The tree-shaking implementations are improving, and so we hope that this situation is getting better soon.

obuchtala:
BTW the warning was there before. Seems to come from rollup.

obuchtala:
I'll merge your commit manually, not via this PR, and add some more of these changes.

obuchtala:
Closing this in favor of #1149

kilburn:
No problem Olivier, thanks for checking it.

After a lot of fiddling around, I ended up budling my own version of the editor based on the simple-writer example (updated to the latest preview version of substance), and using rollup as the bundler. 

Then I just import that bundle as a node package on my main project, where webpack simply doesn't mess up with it.

PR #1155, opened 2018-02-07
Customizable label providers

obuchtala:
In an earlier version of substance we supported i18n via label provider.
Somewhere on the way we have lost this functionality.
This PR is allowing to configure the LabelProvider implementation.
Custom parameters are supported (e.g., for supporting plurals or variables).


obuchtala:
@dotvoid as we talked last time

PR #1156, opened 2018-02-09
Prevent the native behaviour of key combos.

Integral:


ui/Surface.js
```diff
@@ -770,6 +777,16 @@ class Surface extends Component {
     window.setTimeout(fn, BROWSER_DELAY)
   }
 
+  // prevent the native behavior of key combos
+  _muteNativeHandlers(event) {
+    const key = parseKeyEvent(event)
+    const keyboardManager = this.editorSession.keyboardManager
+
+    if(keyboardManager.keydownBindings[key]) {
```
obuchtala: This is maybe a bit implicit. What about adding a switch-case block where we add the keys excplicitly, then everybody understands what we do there.

obuchtala (commented):
In general I think this is the right place to do it.

ui/Surface.js
obuchtala: And I think, we should only suppress default ContentEditable behavior here.
E.g. `CMD-F`  should be suppressed somewhere else.

Integral:
Ah, ok, then it's a different thing.
I mean either we are trying to suppress native behaviour for every
command-related combo, either we are restricting this to number of combos.

obuchtala:
I mean we can add something like what you have done on a higher-level. Just in Surface I'd like to rather suppress only what is 'managed' there, i.e. the ContentEditable.
For App level, we can still decide adding something what you suggested.

obuchtala:
@michael what is ur opinion?

Integral:
Agree.
Still have a question, if we ship find and replace package with substance,
than maybe we should handle this ESC or CMD-f stuff inside that package?

obuchtala:
That would be good. 
At the moment it is not clear to me how we want to generalized this, because we don't have a good answer to the app state yet. (I have to continue my work on that)

PR #1160, opened 2018-02-13
Improvements for schema-driven commands

michael:


ui/InsertInlineNodeCommand.js
```diff
@@ -74,6 +74,10 @@ class InsertInlineNodeCommand extends Command {
     return false
   }
 
+  isAnnotationCommand() {
```
obuchtala: Maybe leave a comment here why you added this.

PR #1175, opened 2018-04-23
Append and remove child API for inline nodes.

Integral:
@oliver---- this is all I need regarding XML inline nodes API.
However there are couple of methods left in XMLElementNode.js which I didn't copied. It maybe a good idea to move them as well for XML API compatibility. What do you think?

### DIFF
--- a/xml/XMLElementNode.js
+++ b/xml/XMLElementNode.js
@@ -1,73 +1,40 @@
 import DOMElement from '../dom/DOMElement'
 import XMLDocumentNode from './XMLDocumentNode'
+import xmlNodeHelpers from './xmlNodeHelpers'
 
 export default
 class XMLElementNode extends XMLDocumentNode {
 
   appendChild(child) {
-    this.insertAt(this._childNodes.length, child)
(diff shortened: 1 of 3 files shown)


obuchtala:
We probably would want to share the implementation by using a set of functions, e.g. `appendChild(xmlNode, newChild)`. Then we can avoid code duplication.

obuchtala:
Would you like to do that? :)

Integral:
Sure!

Integral:
I think you can merge!

obuchtala:
Could you extract the duplicate implementations into a shared file, say `xml/xmlNodeHelpers.js` and use it from `XMLElementNode` and `XMLInlineElementNode`?

Every such shared function would take the `node` as first argument.
For example:
```
appendChild(xmlNode, child) {
  insertAt(xmlNode, xmlNode._childNodes.length, child)
}
```
This way we avoid code duplication.

Integral:
Done!

obuchtala:
Thx!

obuchtala:
Published.

PR #1191, opened 2018-05-14
Fix ES5 build

tamlyn:
As mentioned in #1121, the ES5 build seems to be misconfigured. The `dist/substance.es5.js` file currently published to npm contains `class`, `const`, `let` and other incompatible features. This small fix puts `browser:legacy` builds through Bublé as (I think) intended.

### DIFF
--- a/make.js
+++ b/make.js
@@ -95,7 +95,7 @@ function buildLib(target, production) {
       ]
     }
   }
-  if (target === 'legacy') {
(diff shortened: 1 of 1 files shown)


tamlyn:
Oh, hah. Should have looked at the open PRs.

obuchtala:
Thanks. I try to merge the open PR asap!

PR #1197, opened 2018-05-16
Standard formatting

obuchtala:
Switching from eslint to  standard: https://www.npmjs.com/package/standard
Less to configure than eslint and in Stencila we are using it already.


obuchtala:
@michael @Integral you might need to change the linter setup in your editor. Look here: https://www.npmjs.com/package/standard#are-there-text-editor-plugins

Issue #1116, opened 2017-08-23
Minify problems


Integral:
There is a problems with minifying library.
For instance [this function](https://github.com/substance/substance/blob/develop/model/DocumentChange.js#L44)
turns to
```
_extractInformation(t) {
  function e(e) {
    switch (e.type) {
    case "create":
    case "delete":
      {
        let t = e.val;
        t.hasOwnProperty("start") && (updated[t.start.path] = !0), t.hasOwnProperty("end") && (updated[t.end.path] = !0);
        break
      }
    case "update":
    case "set":
      {
        let n = t.get(e.path[0]);
        n && (n.isPropertyAnnotation() ? updated[n.start.path] = !0 : n.isContainerAnnotation() && affectedContainerAnnos.push(n));
        break
      }
    default:
      throw new Error("Illegal state")
    }
  }
  if (!this._extracted) {
    let n = this.ops, i = {}, r = {}, o = {}, s = [];
    for (let a = 0; a < n.length; a++) {
      let t = n[a];
      "create" === t.type && (i[t.val.id] = t.val, delete r[t.val.id]), "delete" === t.type && (delete i[t.val.id], r[t.val.id] = t.val), "set" !== t.type && "update" !== t.type || (o[t.path] = !0, o[t.path[0]] = !0), e(t)
    }
    s.forEach(function(e) {
      let n = t.get(e.containerId, "strict"), i = n.getPosition(e.start.path[0]), s = n.getPosition(e.end.path[0]);
      for (let a = i; a <= s; a++) {
        let t, e = n.getChildAt(a);
        t = e.isText() ? [e.id, "content"] : [e.id], r[e.id] || (o[t] = !0)
      }
    }), Object.keys(r).length > 0 && _n(o, function(t, e) {
      let n = e.split(",")[0];
      r[n] && delete o[e]
    }), this.created = i, this.deleted = r, this.updated = o, this._extracted = !0
  }
}
```
Which will error as there is no `updated` variable. There are might be an option in uglify to avoid this kind of things, dunno, but it looks like we can't pass any minify option to our bundler.

Integral:
PR: #1117.

obuchtala:
This is a weird bug.
@Integral Is this still happening with the current version? now we use `uglify-es@^3.3.9`

Integral:
[...]

PR #1202, opened 2018-05-18
Minify dist on publish.

Integral:
Additional task to minify `substance.cjs.js`, `substance.es.js` and `substance.js`. As discussed in #1116.

### DIFF
--- a/make.js
+++ b/make.js
@@ -285,7 +285,7 @@ b.task('publish', ['clean', 'lib', 'minify'])
 // Default dev mode, only browser bundles are made and no ES5 transpilation happens
 b.task('dev', ['clean', 'lib:browser:dev', 'test:browser'])
 
-b.task('test', ['test:node', 'cover'])
+b.task('test', ['schema', 'test:node', 'cover'])
   .describe('runs the test suite')
 
 b.setServerPort(4001)
--- a/make.js
+++ b/make.js
@@ -268,12 +268,19 @@ b.task('cover', () => {
   })
 })
 
+b.task('minify', () => {
+  install(b, 'uglify-es', UGLIFY_VERSION)
+  b.minify(DIST + 'substance.cjs.js', DIST + 'substance.cjs.min.js')
+  b.minify(DIST + 'substance.es.js', DIST + 'substance.es.min.js')
+  b.minify(DIST + 'substance.js', DIST + 'substance.min.js')
+})
+
 b.task('vendor', buildVendor)
   .describe('pre-bundles vendor libraries')
 
 b.task('default', ['clean', 'lib'])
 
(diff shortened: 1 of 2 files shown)


obuchtala:
@Integral Just saw that `b.minify()` is not working. I have fixed and published bundler@0.24.2

obuchtala:
Please introduce a 'prepublishOnly' npm script that call `node make publish`, and instead let the `publish` script call `node make`

Integral:
@oliver----  did you mean that `prebublish` script should call `node make`?

obuchtala:
Yeah, that one

obuchtala:
Let's see if the combination of 'prepublish' and 'prepublishOnly' works for us.
Having too much in 'prepublish' is making every 'npm install' so long.

Integral:
Done.

obuchtala:
Thanks! 

Could you try out now if the minified version is working?
For that do a `npm pack` in substance. Then go into texture, and npm install the tarball generated by `npm pack`. Then change `app/index.html` to use `substance.min.js`.

obuchtala:
Oh, I see, `npm pack` is not running `prepublishOnly`... god damnit!

obuchtala:
Could you find out, what script we need, so that `npm pack` is also doing the full build?

obuchtala:
(tststs this npm workflow is anything else but intuitive!)

obuchtala:
After reading https://docs.npmjs.com/misc/scripts
I think we only want `prepack`. Then `node make publish` is run on `npm publish` and `npm pack`, and if you install it via git, i.e. `npm install substance/substance`.
A relative install such as `npm install ../substance` is now pretty much the same as `npm link`. In that case you need to do a manual build anyways.
When I test the bundle, I now always use `npm pack`.

obuchtala:
Sorry for the back-and-forth ;)

obuchtala:
And then a local `npm install` is not doing anything. Which was annoying me anyways.

obuchtala:
@Integral taking over here.. this is bugging me

obuchtala:
This seem now to be doing what I want.
@Integral could you try out too.
Requirements:
- faster local `npm install`s (no need to run build on npm install automatically)
- full build on `npm pack` and `npm publish`
- allow for git builds, i.e, `npm install substance/substance#make-minify`

obuchtala:
Still we need to check if the minified version is working in another project, as a explained above.

obuchtala:
@Integral handing it over to you :)
Thanks!

Integral:
@oliver---- sweet! it's working. Thought I can't really verify that npm installs package from tarball even without internet connection, I could say that minify version working both for app and web versions!

obuchtala:
Ok thanks for checking!

PR #1204, opened 2018-05-26
Fontawesome 5 icon names.

Integral:
If we still want to use font awesome webfonts when we need to update some icon names.


michael:
I think keeping the webfont is fine. I will take a look at your branch in Texture later.

obuchtala:
Obsolete.

PR #1226, opened 2018-09-12
Check if browser is Chromium

Integral:
Sometimes we need to know if the user's browser is Chrome/Chromium.
This PR introduces a platform flag for that.

### DIFF
--- a/util/platform.js
+++ b/util/platform.js
@@ -58,6 +58,7 @@ function detect () {
     const msie = ua.indexOf('MSIE ')
     const trident = ua.indexOf('Trident/')
     const edge = ua.indexOf('Edge/')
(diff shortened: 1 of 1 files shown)


obuchtala:
@Integral the chrome check is not working.
Maybe we should do it more like: https://stackoverflow.com/questions/4565112/javascript-how-to-find-out-if-the-user-browser-is-chrome/13348618#13348618

Integral:
It's worked for me... Let's see tomorrow

obuchtala:
Yes. But as you can see in the linked stackoverflow, just window.chrome is too naive, because it is not the same on different OS, and some browsers like Opera have this set too.

PR #1227, opened 2018-09-13
Better Chrome detection

Integral:
For better Chrome browser detection we want to detect Opera browser and be sure that it's not Opera or EDGE which have similar vendor/user agent properties.


obuchtala:
Thanks!

PR #1259, opened 2019-02-15
Multi view history

obuchtala:
Introducing a change history implementation allowing for multiple simultaneous views.
This is intended for applications having multiple view modes,
that should retain the change history when switching modes.

At the same time this forms a good basis to bring real-time editing further.


michael:
Incredible.

